PR #79, opened 2014-12-19
Single step filter

imh:
At each point in time, we check to see if an observation plausibly comes from the distribution supposed by each hypothesis. If it does not, then we remove the hypothesis from consideration.

### DIFF
--- a/src/ambiguity_test.c
+++ b/src/ambiguity_test.c
@@ -30,6 +30,7 @@
 #define DECORRELATED_PHASE_BIAS_VAR 0
 #define NUM_SEARCH_STDS 5
 #define LOG_PROB_RAT_THRESHOLD -90
+#define SINGLE_OBS_CHISQ_THRESHOLD 20
 
 #define DEBUG_AMBIGUITY_TEST 0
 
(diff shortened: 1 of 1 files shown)


src/ambiguity_test.c
```diff
@@ -485,9 +488,10 @@ void update_and_get_max_ll(void *x_, element_t *elem) {
   for (u8 i=0; i < x->num_dds; i++) {
     hypothesis_N[i] = hyp->N[i];
   }
-
-  hyp->ll += get_quadratic_term(x->res_mtxs, x->num_dds, hypothesis_N, x->r_vec);
+  double q = get_quadratic_term(x->res_mtxs, x->num_dds, hypothesis_N, x->r_vec);
+  hyp->ll += q;
   x->max_ll = MAX(x->max_ll, hyp->ll);
+  return (abs(q) < 20);
```
fnoble: The magic number `20` should probably be a `#define` along with other constants

src/ambiguity_test.c
```diff
@@ -589,7 +593,7 @@ void test_ambiguities(ambiguity_test_t *amb_test, double *dd_measurements)
   x.unanimous_amb_check = &amb_test->amb_check;
   x.unanimous_amb_check->initialized = 0;
 
-  memory_pool_fold(amb_test->pool, (void *) &x, &update_and_get_max_ll);
+  memory_pool_filter(amb_test->pool, (void *) &x, &update_and_get_max_ll);
   /*memory_pool_map(amb_test->pool, &x.num_dds, &print_hyp);*/
```
fnoble: Remove this debugging print code?

fnoble:
Have you tested this to your satisfaction? If so happy to merge after that small change.

fnoble:
Also, can you rebase on top of the current master. Scotts changes are conflicting.

PR #81, opened 2015-01-06
Update tracking loop mathematics and documentation.

gsmcmullin:


gsmcmullin:
The API is compatible with the previous version.  There are some changes in exported functions/structures, but they are not currently being used by piksi_firmware.

The loop dynamics are different, but I believe this is correct.  Most notably, the bandwidth is significantly less after this change.  The response with this change appears to match the input parameters of bandwidth and damping.

PR #85, opened 2015-02-04
Minor fixes required for compiling on MinGW-w64.

jabjoe:
Hi,

Here's my second iteration of this patch after Fergus Nobles change requests.

Joe


fnoble:
Thanks for the patch!

PR #91, opened 2015-03-10
Ephemeris code cleanup

fnoble:


src/ephemeris.c
```diff
@@ -11,144 +12,147 @@
...
 #include "linear_algebra.h"
+#include "constants.h"
 #include "ephemeris.h"
 #include <math.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <assert.h>
 
-#define NAV_OMEGAE_DOT 7.2921151467e-005
-#define NAV_GM 3.986005e14
```
imh: do we not need these anymore?

fnoble: They got moved into `constants.h` to be with their friends

src/ephemeris.c
```diff
@@ -11,144 +12,147 @@
...
-int calc_sat_pos(double pos[3], double vel[3],
-             double *clock_err, double *clock_rate_err,
-             const ephemeris_t *ephemeris,
-             gps_time_t tot)
+/** Calculate satellite position, velocity and clock offset from ephemeris.
+ *
+ * References:
+ *   -# IS-GPS-200D, Section 20.3.3.3.3.1 and Table 20-IV
+ *
+ * \param t GPS time at which to calculate the satellite parameters
```
imh: nit: If the function is called "calc_sat_state" we might want to say "...at which to calculate the satellite state" instead of "...at which to calculate the satellite parameters" for consistency.

src/ephemeris.c
```diff
@@ -156,27 +160,29 @@ double predict_range(double rx_pos[3],
...
-/** Do we have an ephemeris for this PRN?
- * \TODO  This should actually be more than just the "valid" flag.
- *         When we write an is_usable() function, lets use that instead
- *         of just es[prn].valid.
+/** Is this ephemeris usable?
  *
- * \param prn     The prn we want to check.
+ * \todo This should actually be more than just the "valid" flag.
+ *       When we write an is_usable() function, lets use that instead
+ *       of just es[prn].valid.
```
imh: In a PR of mine somewhere, there's an added "healthy" flag. Might be bundled in with the sat set handling changes that fixed the LD issue..

imh:
Looks good

src/ephemeris.c
fnoble: I'll look out for that when we merge your PR

PR #92, opened 2015-03-10
Logging subsystem

fnoble:
WIP


include/libswiftnav/logging.h
```diff
@@ -0,0 +1,84 @@
...
+ *
+ *    #define DEBUG 1
+ *
+ * \{ */
+
+/** Log an error.
+ * \param args `printf` style format and arguments.
+ */
+#define log_error(args...) printf("ERROR: " args)
+
```
denniszollo: I think perhaps we should make these macro functions all caps, as to help indicate to the code reader that they are macros. The Kernel standards say this is optional, but I like to do it. 
excerpt from kernel coding standards
 "CAPITALIZED macro names are appreciated but macros resembling functions
may be named in lower case."

fnoble: Usually I would agree, but I think there is a good chance they may not be macros in the final implementation...

src/amb_kf.c
```diff
@@ -22,14 +22,16 @@
...
+#include "logging.h"
+#include "linear_algebra.h"
 #include "constants.h"
 #include "track.h"
 #include "almanac.h"
 #include "gpstime.h"
 #include "amb_kf.h"
 
-#define DEBUG_AMB_KF 0
+#define DEBUG 0
```
denniszollo: Don't we want to define DEBUG before including logging.h? If I change this debug to 1, I think the compiler will warm me about multiple definitions of the macro.

gsmcmullin:
Looks good.  A big improvement.

PR #103, opened 2015-03-20
Low latency solution checks if the KF has enough sats

imh:
Also added some asserts, log_debugs, and documentation for it.
(low latency fixed already had the necessary checks)

### DIFF
--- a/src/dgnss_management.c
+++ b/src/dgnss_management.c
@@ -12,6 +12,7 @@
 
 #include <string.h>
 #include <stdio.h>
+#include <assert.h>
 #include "logging.h"
 #include "amb_kf.h"
 #include "stupid_filter.h"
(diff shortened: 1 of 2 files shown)


src/dgnss_management.c
```diff
@@ -561,6 +565,16 @@ s8 dgnss_low_latency_baseline(u8 num_sdiffs, sdiff_t *sdiffs,
                               double ref_ecef[3], u8 *num_used, double b[3])
 {
   DEBUG_ENTRY();
+  if (num_sdiffs < 4 || sats_management.num_sats < 4) {
+    /* For a position solution, we need at least 4 sats. That means we must
+     * have at least 4 sats in common between what the filters are tracking and
+     * the sdiffs we give this function. If num_sdiffs, or the number of KF
+     * sats is less than 4, this criterion cannot be satisfied. */
+    log_debug("Low latency solution can't be computed. Too few observations"
```
fnoble: I think this should be a `log_error`, this is was that macro is for!

imh: Not in this case. We expect this to happen all the time, this is just an early return condition, not a bad error.

imh: oh wait, i see what you meant. derp

imh:
jenkins pleast test this

imh:
retest this please

PR #105, opened 2015-03-23
Remove duplicate UDU/eye/triu functions from lambda.c

fnoble:
cc: @imh

### DIFF
--- a/src/lambda.c
+++ b/src/lambda.c
@@ -19,6 +19,7 @@
 #include <cblas.h>
 #include <clapack.h>
 
+#include "linear_algebra.h"
 #include "amb_kf.h"
 #include "logging.h"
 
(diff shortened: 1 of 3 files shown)


src/lambda.c
```diff
@@ -30,52 +31,6 @@
...
-static s8 udu2(u32 n, double *M, double *U, double *D) //todo: replace with DSYTRF
-{
-  double alpha, beta;
-  triu2(n, M);
-  eye2(n, U);
-  memset(D, 0, n * sizeof(double));
-
-  for (u32 j=n; j>=2; j--) {
-    D[j - 1] = M[(j-1)*n + j-1];
-    if (D[j-1] > 0) {
```
imh: The matrix_udu function to replace this has "D[j-1] != 0" instead of "D[j-1] > 0" here. Before I sign off, I want to check which one stays positive semidefinite when we get really close to singular.

imh:
since you made those tests, lgtm

PR #106, opened 2015-03-24
Initializing floating sat amb estimates with code + carrier

imh:


src/amb_kf.c
```diff
@@ -35,6 +35,10 @@
  * The expectation value of carrier + code / lambda is
  * integer ambiguity + bias. Regardless of bias, this is an
  * important measurement.
+ *
+ * \param carrier A carrier phase measurement in units of wavelengths.
+ * \param code    A code measurement in the same units as GPS_L1_LAMBDA_NO_VAC.
+ * \return A rough estimate of the integer ambiguity.
```
fnoble: rough here sounds like imprecise language, perhaps consider dropping that and detailing the limitations of this measurement technique above?

src/amb_kf.c
```diff
@@ -853,11 +857,30 @@ void nkf_state_projection(nkf_t *kf,
   /* NOTE: IT DOESN'T UPDATE THE OBSERVATION OR TRANSITION MATRICES, JUST THE STATE. */
 }
 
+/** Add new sats the the KF
```
fnoble: s/the/to/
KF -> Kalman filter

src/amb_kf.c
```diff
@@ -853,11 +857,30 @@ void nkf_state_projection(nkf_t *kf,
...
+ *
+ * \param kf                    The KF struct to be updated
+ * \param num_old_non_ref_sats  The old number of DDs (dimension of X)
+ * \param num_new_non_ref_sats  The new number of DDs (dimension of Z)
+ * \param ndx_of_old_sat_in_new The indices of the old sats in the new full
+ *                              space. For example, if the first (ndx=0) element
+ *                              of the state vector in X was for PRN 3 and
+ *                              PRN 3 is the 4th (ndx=3) element of the state
+ *                              vector in Z, then ndx_of_old_sat_in_new[0] = 3.
+ * \param estimates             Rough estimates of the ambiguities (in Z).
```
fnoble: Again drop rough and consider expanding on the fact that this should be length `num_new_non_ref_sats`, and is used to initialise the starting value of the state mean for the new satellites.

src/amb_kf.c
```diff
@@ -853,11 +857,30 @@ void nkf_state_projection(nkf_t *kf,
...
+ * \param int_init_var          The variance with which to initialize any new
```
fnoble: Maybe make clear that this is the variance associated with `estimates`

src/amb_kf.c
fnoble: Also, what about renaming `estimates` to `init_amb_est` or something to make it more self documenting what the estimates are of. Same goes for the other uses of `estimates`

src/amb_kf.c
```diff
@@ -867,7 +890,7 @@ void nkf_state_inclusion(nkf_t *kf,
   double new_cov[new_state_dim * new_state_dim];
   memset(new_cov, 0, new_state_dim * new_state_dim * sizeof(double));
   double new_mean[new_state_dim];
-  memset(new_mean, 0, new_state_dim * sizeof(double));
+  memcpy(new_mean, estimates, new_state_dim * sizeof(double));
```
fnoble: Maybe a comment to say that you are initializing the means here with the ambiguity estimates, and the means for the non new satellites will be overwritten with their existing values below.

src/dgnss_management.c
```diff
@@ -160,6 +160,24 @@ void sdiffs_to_prns(u8 n, sdiff_t *sdiffs, u8 *prns)
   }
 }
 
+/** Roughly measure the ambiguity vector given sdiffs.
```
fnoble: Jus' hating on "roughly" today

include/libswiftnav/dgnss_management.h
```diff
@@ -22,7 +22,7 @@
 #define DEFAULT_CODE_VAR_KF     (100 * 400)
 #define DEFAULT_AMB_DRIFT_VAR   1e-8
 #define DEFAULT_AMB_INIT_VAR    1e8
-#define DEFAULT_NEW_INT_VAR     1e10
+#define DEFAULT_NEW_INT_VAR     1e4
```
denniszollo: Perhaps we can put some comments to describe the different initial values?  I don't really understand the difference between "initial" and "new".  Also, these numbers are settings currently in the firmware (whose default value is this preprocesser macro).  It's unrelated to this PR, but should they remain settings as we move towards 1.0?

imh: Done. The difference between initial and new should be deprecated soon. Now that we aren't initializing and adding sats as naively, we use the same estimator to start the KF as we do to add sats. So these variance parameters will be the same.

(But before I deprecate it, I need to make sure 1e4 is a good value for initialization too).

PR #113, opened 2015-04-03
Generate histogram for finding nav bit edge.

gsmcmullin:
Fixes swift-nav/piksi_firmware#305.

### DIFF
--- a/src/nav_msg.c
+++ b/src/nav_msg.c
@@ -19,7 +19,7 @@
 #include "bits.h"
 #include "nav_msg.h"
 
-#define NAV_MSG_BIT_PHASE_THRES 5
+#define NAV_MSG_BIT_PHASE_THRES 10
 
 void nav_msg_init(nav_msg_t *n)
 {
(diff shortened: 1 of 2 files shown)


src/nav_msg.c
```diff
@@ -73,6 +73,36 @@ static u32 extract_word(nav_msg_t *n, u16 bit_index, u8 n_bits, u8 invert)
...
+  float dot = corr_prompt_real * n->prev_corr;
+  n->prev_corr = corr_prompt_real;
+
+  if (dot > 0)
+    return;
+
+  /* Sign change - Add to histogram */
+  n->hist[n->bit_phase] += -dot;
+
+  n->bit_phase_count++;
```
fnoble: Add comment to explain this waits until we have seen `NAV_MSG_BIT_PHASE_THRES` sign flips before looking at the histogram

fnoble:
Looks good as is

Created #115 to keep track of some further improvements we could make in the future.

PR #114, opened 2015-04-03
Increase the size of nav message subframe bit buffer and check for overrun.

gsmcmullin:
This fixes a lot of 'subframe parity error' messages when the CPU load increases and the nav msg thread is unable to process the subframe in time.  This should also fix the problem of occasionally getting a bad ephemeris.

### DIFF
--- a/include/libswiftnav/nav_msg.h
+++ b/include/libswiftnav/nav_msg.h
@@ -16,11 +16,12 @@
 #include "common.h"
 #include "ephemeris.h"
 
-#define NAV_MSG_SUBFRAME_BITS_LEN 12 /* Buffer 384 nav bits. */
+#define NAV_MSG_SUBFRAME_BITS_LEN 14 /* Buffer 448 nav bits. */
 
 typedef struct {
   u32 subframe_bits[NAV_MSG_SUBFRAME_BITS_LEN];
   u16 subframe_bit_index;
+  bool overrun;
   /** Foo
    * - 0 = no preamble found
    * - +x = preamble begins at bit index (x-1)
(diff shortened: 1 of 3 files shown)


src/nav_msg.c
```diff
@@ -114,6 +114,20 @@ s32 nav_msg_update(nav_msg_t *n, s32 corr_prompt_real, u8 ms)
   /* Dump the nav bit, i.e. determine the sign of the correlation over the
    * nav bit period. */
 
+#define abs(x) ((x) < 0 ? -(x) : (x))
```
fnoble: Either use the one from `math.h` or move to `common.h`

include/libswiftnav/nav_msg.h
```diff
@@ -16,11 +16,12 @@
 #include "common.h"
 #include "ephemeris.h"
 
-#define NAV_MSG_SUBFRAME_BITS_LEN 12 /* Buffer 384 nav bits. */
+#define NAV_MSG_SUBFRAME_BITS_LEN 14 /* Buffer 448 nav bits. */
```
fnoble: Feel free to increase further if you think it would be beneficial. Obviously we should fix the priorities as well.

src/nav_msg.c
```diff
@@ -114,6 +114,20 @@ s32 nav_msg_update(nav_msg_t *n, s32 corr_prompt_real, u8 ms)
...
+  u16 last_subframe_bit_index = abs(n->subframe_start_index) + 27;
```
fnoble: Maybe a comment explaining the significance of 27 here?

fnoble:
Looks good

PR #117, opened 2015-04-11
PVT altitude sanity check: halfway to the moon isn't very sane

henryhallam:
Closes swift-nav/piksi_firmware#339

### DIFF
--- a/src/pvt.c
+++ b/src/pvt.c
@@ -260,10 +260,10 @@ static double pvt_solve(double rx_state[],
 u8 filter_solution(gnss_solution* soln, dops_t* dops)
 {
   if (dops->pdop > 50.0)
-    /* PDOP is too high to yeild a good solution. */
+    /* PDOP is too high to yield a good solution. */
     return 1;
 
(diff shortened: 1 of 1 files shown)


fnoble:
Hmm.. `soln_llh` is in meters internally so this should be correct. Are you sure you verified the bug?

henryhallam:
@fnoble per [pvt.h](https://github.com/swift-nav/libswiftnav/blob/master/include/libswiftnav/pvt.h#L36) it's in meters.  What I saw was very consistent with the solution being dropped below -1 m.  I feel like putting it on the signal simulator would be a waste of time, but I'll do it if you want.

henryhallam:
Derp, I mean yes I see what you mean. Hmm.

henryhallam:
..in which case 1e8 seems too high for the upper limit, given that's halfway to the moon.

henryhallam:
We saw Piksi outputting solutions with altitude in the ~ -10 m range, so clearly this was bogus as originally filed.

PR #136, opened 2015-04-28
Experimental change to select shortest pseudo range for ordinal satellite

cturvey:
Modified calc_navigation_measurement_() to select the shortest path satellite for the ordinal satellite.

Current code picks minimum ToW (earlier) for dispatch time of signal vs current ToW at receiver. Should perhaps be the minimum delta in this context, not min/earliest time. This results in it picking the furthest satellite, lowest, with the worst signal path, and also prone to fade in/out. The observation (Ken) is that the RTK solution jumps in these cases, and you are also likely to see the biggest change in receiver clock drift/bias, which I'd otherwise expect to have a reasonably consistent rate-of-change. Haven't dug deep enough to see where receiver drift/bias are fixed, or if they are filtered.

The shortest path would tend to favour the highest elevation, clearest view satellite to be selected as the ordinal satellite, ie the one with the 2298000000 arbitrary measurement. This is the only place this constant is utilized.

-Clive

### DIFF
--- a/src/track.c
+++ b/src/track.c
@@ -652,8 +652,10 @@ void calc_navigation_measurement_(u8 n_channels, channel_measurement_t* meas[], 
 {
   double TOTs[n_channels];
   double min_TOT = DBL_MAX;
+  double max_TOT = DBL_MIN;
 
   for (u8 i=0; i<n_channels; i++) {
+    /** \todo Optimize the math here, with delta, and reciprocals */
     TOTs[i] = 1e-3 * meas[i]->time_of_week_ms;
     TOTs[i] += meas[i]->code_phase_chips / 1.023e6;
     TOTs[i] += (nav_time - meas[i]->receiver_time) * meas[i]->code_phase_rate / 1.023e6;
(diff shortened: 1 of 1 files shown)


fnoble:
ok to test

fnoble:
Thanks Clive!

cturvey:
No problem.

Where about in the code should I look for the fixing of the receivers clock
drift/bias? Does it remember or filter this epoch to epoch? Does it keep a
running value for clock drift, and does it slew it periodically?

fnoble:
We don't filter or store the clock drift right now as we currently don't have that much use for it. You might be able to take a look at the `clock_offset` and `clock_bias` members of the `gnss_solution` output by `calc_PVT`. They contain the offset between the assumed clock implied by the choice of reference and GPS time, and the clock frequency bias.

meee1:
ive added another pull request that I believe is more accurate than this one.

fnoble:
A derivative of this was merged in #138.

PR #137, opened 2015-04-30
swig interface files for libswiftnav.

mookerji:
Our bindings to libswiftnav get stale pretty quickly, and lack full,
or even accurate, coverage of some core functionality. While working
with libswiftnav-python for the filter testing, I'm also experimenting
with some Swig-generated bindings in parallel so we can reuse some
more work going forward.

This includes common interface definitions using only the header
files. libswiftnav-python will consume these from libswiftnav via a
submodule.

/cc @cbeighley @fnoble @mfine

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,6 @@
 build*/
 plover/dist/
 *.pyc
-sbp_generate/sbp_out.*
 clapack-3.2.1-CMAKE/F2CLIBS/libf2c/arith*
+swig/*.c
+swig/*.py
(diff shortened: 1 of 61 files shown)


mookerji:
add to whitelist

henryhallam:
Missing newlines at the end of all files

fnoble:
1) I think the bindings should be build by libswiftnav and the libswiftnav-python package can finally die (CMake already has good support for swig)
2) Have you tried using any of the bindings? My fear is anything that takes an array right now might not work

mookerji:
1) Great idea. Hold off on merging this until then.
2) Good point.

mookerji:
see: 
https://github.com/swift-nav/libswiftnav/pull/137/files#diff-105e764a5ea4c7adfd3b232b4bb802f8R14

The array use case here is a bit problematic, which can be dealt with using Python's C extension API.
https://gist.github.com/mookerji/a74db439d59bfdc205ad

JoshuaGross:
Conflicts and this is old. Can we close this?

PR #138, opened 2015-04-30
use High elevation sat for nominal range

meee1:
this is a different version of
https://github.com/swift-nav/libswiftnav/pull/136
and believe this version is more correct, because each tot is corrected for satellite bias, before making the selection

### DIFF
--- a/src/track.c
+++ b/src/track.c
@@ -651,7 +651,8 @@ void calc_navigation_measurement(u8 n_channels, channel_measurement_t meas[],
 void calc_navigation_measurement_(u8 n_channels, channel_measurement_t* meas[], navigation_measurement_t* nav_meas[], double nav_time, ephemeris_t* ephemerides[])
 {
   double TOTs[n_channels];
-  double min_TOT = DBL_MAX;
+  double min_TOF = -DBL_MAX;
+  double clock_err[n_channels], clock_rate_err[n_channels];
 
   for (u8 i=0; i<n_channels; i++) {
     TOTs[i] = 1e-3 * meas[i]->time_of_week_ms;
@@ -665,9 +666,6 @@ void calc_navigation_measurement_(u8 n_channels, channel_measurement_t* meas[], 
     if (gpsdifftime(nav_meas[i]->tot, ephemerides[i]->toe) > 3*24*3600)
       nav_meas[i]->tot.wn -= 1;
 
-    if (TOTs[i] < min_TOT)
-      min_TOT = TOTs[i];
-
     nav_meas[i]->raw_doppler = meas[i]->carrier_freq;
     nav_meas[i]->snr = meas[i]->snr;
     nav_meas[i]->prn = meas[i]->prn;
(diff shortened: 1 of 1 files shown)


cturvey:
From static analysis it appears to be computationally heavy (2x), and results in the ordinal satellite to report with non GPS_NOMINAL_RANGE value.

The purpose of the earlier patch was to demo higher elevation satellites over gutter balls.

It's not apparent that down stream code filters, or uses as a starting point, the current clock bias (accumulated) and drift (rate of change), rather is recomputes iteratively the receiver clock offset for each set of observation. Other receiver designs would use the clock drift of the receivers to suitably adjust the advance of the correlators, and push the doppler search windows based on the actual clock being used to down convert to the IF (carrier) and the local code generation (code).

cturvey:
On further review looks to retain GPS_NOMINAL_RANGE value

src/track.c
```diff
@@ -676,12 +675,18 @@ void calc_navigation_measurement_(u8 n_channels, channel_measurement_t* meas[],
     nav_meas[i]->carrier_phase += (nav_time - meas[i]->receiver_time) * meas[i]->carrier_freq;
 
     nav_meas[i]->lock_counter = meas[i]->lock_counter;
+	
+	// calc the clock error
+	calc_sat_state(ephemerides[i], nav_meas[i]->tot,
+                   nav_meas[i]->sat_pos, nav_meas[i]->sat_vel,
+                   &clock_err, &clock_rate_err);
```
fnoble: Do you think this could be refactored to call `calc_sat_state` only once? Its a very expensive call, accounts for a good fraction of the PVT solution time.

fnoble: Also, I believe the satellite clock error is small enough (<1% of pseudorange) that this effect isn't very significant, but if its possible to do it this way without additional overhead then it sounds good to me.

cturvey: I think you have look algorithmically outside of this code as to the effect
of selecting and remaining on a specific ordinal satellite. Not sure I saw
any elevation masking code, or lets stick with this satellite until it
reaches X degrees of elevation, type code.

I targeted finding a suitably high one so the transition between visible
satellites has the least impact on the numbers, rather than explicitly
defining that it needs to be the *highest*.

Ken's observation is that all the RTK solution jumping (decamping) is
around the time the satellite which currently gets picked fades in and out
of visibility and lock. Mine more generally is that rate of change (clock
drift) on the receiver is generally fairly consistent short of someone
blowing on the XO/TCXO (which causes massive acceleration/deceleration
events to befall the correlators). I'm bringing my own baggage to the party
here, as I don't have a sufficiently clear view of the software/hardware in
this case. I'll talk with Fergus/Colin about this in another venue. My
other observation is there is a lot of math here that's done over and over,
so there's less opportunity to view an epoch-to-epoch change.

I'm tinkering with other changes to separate the received observations from
the common set used for the RTK solution. The buffers for RTK or matrix
rise to the square so you're pretty much memory locked to 11 channels
(perhaps eek 12), when realistically you could get 18 measurements from a
base (sub 256 byte packet, not sure I ever see more than 14), and get down
to 8 or so good candidates for the RTK side, ie good sustained cycle lock
and higher mask angle.

cturvey:
More concisely I think that the receiver clock offset from epoch-to-epoch depends not only on the clock performance of the receiver, but also the motion/speed (doppler) of the chosen ordinal satellite, mixed to some degree with the others, and abrupt epoch-to-epoch changes of rate, and clock offset, when the role of ordinal satellite transitions to another, or ping-pongs back and forth due to visibility/masking concerns.

My sense is that this is why the post processing stuff hates the observations, as they expect a linear epoch-to-epoch clock drift they can compute and apply over the measurements, and instead we have something more quadratic in nature where every epoch get resolved with two variables.

I'll note that the RINEX data I see from LEICA L1+L2+GLONASS type receivers is that the pseudo-range, carrier and doppler are all solved cleanly, ie you can take the carrier cycles, wavelengths and get the range.

meee1:
ive just refactored this to be more cpu friendly

meee1:
here is the other type of refactor
https://github.com/meee1/libswiftnav/commit/afd8de372ae9a503f325d7d91e571ae7d5d6f67a

fnoble:
ok to test

fnoble:
@meee1 : I like the second option (https://github.com/meee1/libswiftnav/commit/bc55bf53ccb1a216310408e7c10e80260ff6f595) as it keeps the clock error calculation encapsulated. We can spare the extra memory.

fnoble:
@cturvey : I agree we want to use a proper clock model at some point so the pseudoranges become closer to being ranges. Until then I think your fix is a big improvement.

I don't feel strongly about which one of these PRs to proceed with, as I don't think there is going to be much performance difference between them.

@meee1 :  Could you update to use the v2 method and fix the indentation and comment style (we use the C `/* */` style), then we can start moving this through QA

meee1:
@fnoble pull request updated as requested

src/track.c
```diff
@@ -676,22 +674,25 @@ void calc_navigation_measurement_(u8 n_channels, channel_measurement_t* meas[],
...
-  }
+	
+    /* calc sat clock error */
+    calc_sat_state(ephemerides[i], nav_meas[i]->tot,
+                   nav_meas[i]->sat_pos, nav_meas[i]->sat_vel,
+                   &clock_err[i], &clock_rate_err[i]);
 
-  double clock_err, clock_rate_err;
+    /* remove clock error to put all tots within the same time window */
+    if ((TOTs[i] + clock_err[i]) > min_TOF)
```
meee1: there is still a chance that this condition could pick the seconds highest

cbeighley:
@fnoble @henryhallam please review so I can merge

fnoble:
lgtm

fnoble:
Thanks @meee1 and @cturvey !

PR #148, opened 2015-05-13
Misc refactoring + unit tests

fnoble:
- Create `filter_utils.c`
 - Move some common code into `filter_utils`
 - Add some unit tests
 - Clean up unit test output (shouldn't be calling `printf` all over the place)

cc: @imh @henryhallam


imh:
looks great!

PR #149, opened 2015-05-14
Baseline Refactor

fnoble:


fnoble:
@kovach @cbeighley @imh : monster PR ready for review :)

imh:
renamine single_diff to observation seems ambiguous. Maybe we can come up with a better name, since single differenced, double differenced, and undifferenced observations are all combinations of each other, and the single diff is only special as the primitive dual receiver observation.

fnoble:
The thought was to put all the single, double and un-differenced observation related stuff into one file. `single_diff.c` was starting to contain more than just single diff related code.

Was thinking of moving `navigation_measurement_t` in here too at some point.

imh:
A lot of these commits overlap and would have been easier to review if they'd bee squashed together. Squashing them would have been easier if this PR had been a smaller focused change.

imh:
One thing we talked about months ago when we talked about a long time ago when we were thinking about the set stuff was whether our intersection function should intersect distinct types. The alternative is that the things to be intersected both have some kind of same-typed key. Doing it with keys only requires n comparator functions for n types, as opposed to n^2. But it requires the overhead of using key functions. We'll see if it ends up hairy.

fnoble:
@imh : code review comments addressed now

PR #150, opened 2015-05-15
REFACTOR: Remove global dgnss state; add some const-correctness

henryhallam:
The structs settings, nkf, sats_management and ambiguity_test
which maintain the state of the DGNSS filters were previously
globals in dgnss_management.c.  Now they are members of the context
struct 'dgnss_state_t' which should be instantiated in the host program
and initialized via dgnss_init().  A pointer to this struct must be
passed to most dgnss_* functions.

Many functions are now also const-correct, though I have not attempted
complete const-correct coverage.

Changes to piksi_firmware, libswiftnav-python and gnss_analysis will be
required.

/cc @fnoble @imh for review, @mookerji FYI

### DIFF
--- a/include/libswiftnav/amb_kf.h
+++ b/include/libswiftnav/amb_kf.h
@@ -40,9 +40,9 @@ void nkf_update(nkf_t *kf, double *measurements);
 
 void assign_phase_obs_null_basis(u8 num_dds, double *DE_mtx, double *q);
 void set_nkf(nkf_t *kf, double amb_drift_var, double phase_var, double code_var, double amb_init_var,
-            u8 num_sdiffs, sdiff_t *sdiffs_with_ref_first, double *dd_measurements, double ref_ecef[3]);
+            u8 num_sdiffs, const sdiff_t *sdiffs_with_ref_first, const double *dd_measurements, const double ref_ecef[3]);
 void set_nkf_matrices(nkf_t *kf, double phase_var, double code_var,
-                     u8 num_sdiffs, sdiff_t *sdiffs_with_ref_first, double ref_ecef[3]);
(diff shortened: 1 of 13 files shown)


henryhallam:
ok I need to work on the unit tests

include/libswiftnav/amb_kf.h
```diff
@@ -40,9 +40,9 @@ void nkf_update(nkf_t *kf, double *measurements);
 
 void assign_phase_obs_null_basis(u8 num_dds, double *DE_mtx, double *q);
 void set_nkf(nkf_t *kf, double amb_drift_var, double phase_var, double code_var, double amb_init_var,
-            u8 num_sdiffs, sdiff_t *sdiffs_with_ref_first, double *dd_measurements, double ref_ecef[3]);
+            u8 num_sdiffs, const sdiff_t *sdiffs_with_ref_first, const double *dd_measurements, const double ref_ecef[3]);
```
mookerji: **great**

mookerji: +22

fnoble:
lgtm once its tested

src/dgnss_management.c
```diff
@@ -64,10 +36,10 @@ void make_measurements(u8 num_double_diffs, const sdiff_t *sdiffs, double *raw_m
   DEBUG_EXIT();
 }
 
-static bool prns_match(const u8 *old_non_ref_prns, u8 num_non_ref_sdiffs,
+static bool prns_match(const dgnss_state_t *dgs, const u8 *old_non_ref_prns, u8 num_non_ref_sdiffs,
```
imh: this doesn't need the whole state, just the sats management. Why not pass that in instead?

imh:
There are a whole bunch of places where we are operating on the KF, or on sats man, but pass the whole state struct. Why is that?

PR #167, opened 2015-06-10
fixed multiple off by one bugs

kovach:
fixes an issue that may cause memory corruption when the ref satellite changes and the old ref is the largest prn in the sats_management struct

### DIFF
--- a/src/amb_kf.c
+++ b/src/amb_kf.c
@@ -669,6 +669,13 @@ void set_nkf_matrices(nkf_t *kf, double phase_var, double code_var,
                   kf->decor_obs_mtx);
 }
 
+/** Currently this function is only used to find the index of a prn in a
+ *  permuted basis, so all callsites assert the result is not -1.
+ *  \param num_elements Size of list
+ *  \param x the prn to search for
+ *  \param list the prn list to search
+ *  \return index of x in list, or -1 if no element is equal to x
+ */
 s32 find_index_of_element_in_u8s(const u32 num_elements, const u8 x, const u8 *list)
 {
   for (u32 i=0; i<num_elements; i++) {
@@ -688,8 +695,15 @@ void rebase_mean_N(double *mean, const u8 num_sats, const u8 *old_prns, const u8
   u8 old_ref = old_prns[0];
   u8 new_ref = new_prns[0];
 
+  if (old_ref == new_ref) {
+    /* Nothing needs to be done; same basis. */
+    return;
+  }
+
   double new_mean[state_dim];
-  s32 index_of_new_ref_in_old = find_index_of_element_in_u8s(num_sats, new_ref, &old_prns[1]);
+  s32 index_of_new_ref_in_old = find_index_of_element_in_u8s(num_sats-1, new_ref, &old_prns[1]);
+  assert(index_of_new_ref_in_old != -1);
+
   double val_for_new_ref_in_old_basis = mean[index_of_new_ref_in_old];
   for (u8 i=0; i<state_dim; i++) {
     u8 new_prn = new_prns[1+i];
@@ -697,7 +711,8 @@ void rebase_mean_N(double *mean, const u8 num_sats, const u8 *old_prns, const u8
       new_mean[i] = - val_for_new_ref_in_old_basis;
     }
     else {
-      s32 index_of_this_sat_in_old_basis = find_index_of_element_in_u8s(num_sats, new_prn, &old_prns[1]);
+      s32 index_of_this_sat_in_old_basis = find_index_of_element_in_u8s(num_sats-1, new_prn, &old_prns[1]);
+      assert(index_of_this_sat_in_old_basis != -1);
       new_mean[i] = mean[index_of_this_sat_in_old_basis] - val_for_new_ref_in_old_basis;
     }
   }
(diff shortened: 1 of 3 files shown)


src/sats_management.c
```diff
@@ -142,11 +145,14 @@ static void set_reference_sat(const u8 ref_prn, sats_management_t *sats_manageme
           j++;
         }
       }
-      else if (i == sats_management->num_sats-1) {
-        sats_management->prns[j] = old_ref;
-      }
```
kovach: shouldn't have been guarded by 'else'

src/sats_management.c
```diff
@@ -240,7 +246,7 @@ s8 rebase_sats_management(sats_management_t *sats_management,
   }
   else {
     sdiff_t intersection_sats[num_sdiffs];
-    u8 num_intersection = intersect_sats(sats_management->num_sats, num_sdiffs,
+    u8 num_intersection = intersect_sats(sats_management->num_sats-1, num_sdiffs,
                                          &(sats_management->prns[1]), sdiffs, intersection_sats);
```
kovach: reads past end of prns list

src/sats_management.c
```diff
@@ -142,11 +145,14 @@ static void set_reference_sat(const u8 ref_prn, sats_management_t *sats_manageme
...
       }
-      else if (i == sats_management->num_sats-1) {
-        sats_management->prns[j] = old_ref;
-      }
     }
+    if (set_old_yet == 0) {
+      or_count++;
+      sats_management->prns[j] = old_ref;
+    }
+    assert (or_count == 1);
```
kovach: I maybe should take this assertion out

src/amb_kf.c
```diff
@@ -531,7 +531,7 @@ s32 find_index_of_element_in_u8s(const u32 num_elements, const u8 x, const u8 *l
       return i;
     }
   }
-  return -1;
+  assert(false);
```
fnoble: I'd leave this function returning `-1` for error and check the return code where its used. Also would be good to add a doxygen comment for the function and explaining the return value.

src/amb_kf.c
```diff
@@ -544,7 +544,7 @@ void rebase_mean_N(double *mean, const u8 num_sats, const u8 *old_prns, const u8
   u8 new_ref = new_prns[0];
 
   double new_mean[state_dim];
-  s32 index_of_new_ref_in_old = find_index_of_element_in_u8s(num_sats, new_ref, &old_prns[1]);
+  s32 index_of_new_ref_in_old = find_index_of_element_in_u8s(num_sats-1, new_ref, &old_prns[1]);
```
fnoble: Should we use `state_dim` for `num_sats-1`? At least we should be consistent.

src/sats_management.c
fnoble: `or_count` is just a check that we insert the old ref only once right? If its no longer useful then take it out otherwise if you think its a good cross-check then it can stay. I don't mind.

src/sats_management.c
```diff
@@ -129,9 +130,11 @@ static void set_reference_sat(const u8 ref_prn, sats_management_t *sats_manageme
     memcpy(old_prns, sats_management->prns, sats_management->num_sats * sizeof(u8));
```
fnoble: Could you add a few comments to this function for future explorers? I feel like this may not be the last time this function gets revisited...

fnoble:
Couple comments and if we allow the `-1` return code from the index finding function when we should also check the return code all the places its used. Otherwise looks good!

src/sats_management.c
kovach: TODO fix function just above this one

src/amb_kf.c
kovach: I think num_sats - 1 is the length of &old_prns[1], but the mean is length state_dim

src/amb_kf.c
kovach: maybe change the name and add comment to indicate it should only be used on permuted prn lists, keep assert(false)?

src/amb_kf.c
fnoble: Line 541 says `u8 state_dim = num_sats - 1;`?

src/amb_kf.c
fnoble: I feel like this is a pretty obvious way the function can fail and an assert may be a bit heavy handed here. I think its best to let the caller handle it explicitly. I see the point that sometimes they don't and we get bugs but I'm not sure that making it a fatal error is the best solution. I think we should encourage the caller to be more diligent.

src/amb_kf.c
```diff
@@ -524,6 +524,8 @@ void set_nkf_matrices(nkf_t *kf, double phase_var, double code_var,
                   kf->decor_obs_mtx);
 }
 
+/* Currently this function is only used to find the index of a prn in a
+ * permuted basis, so all callsites assert the result is not -1. */
```
fnoble: Turn this into a Doxygen style comment with description of parameters and return values. There is an example of the formatting in the eng. playbook.

src/sats_management.c
```diff
@@ -91,29 +94,31 @@ void set_reference_sat_of_prns(const u8 ref_prn, const u8 num_sats, u8 *prns)
...
       }
-      else if (i == num_sats-1) {
-        prns[j] = old_ref;
-      }
+    }
+    if (set_old_yet == 0) {
+      prns[j] = old_ref;
+      set_old_yet++;
     }
   }
```
fnoble: Might be a good place for an `assert(set_old_yet == 1);`

src/sats_management.c
```diff
@@ -79,8 +80,10 @@ static u8 intersect_sats(const u8 num_sats1, const u8 num_sdiffs, const u8 *sats
   return n;
 }
 
-/** Puts sdiffs into sdiffs_with_ref_first with the sdiff for ref_prn first
+/** Puts sdiffs into sdiffs_with_ref_first with the sdiff for ref_prn first.
```
fnoble: This function doesn't seem to have anything to do with sdiffs?

src/sats_management.c
```diff
@@ -129,24 +134,25 @@ static void set_reference_sat(const u8 ref_prn, sats_management_t *sats_manageme
...
       }
-      else if (i == sats_management->num_sats-1) {
-        sats_management->prns[j] = old_ref;
-      }
+    }
+    if (set_old_yet == 0) {
+      sats_management->prns[j] = old_ref;
+      set_old_yet++;
     }
   }
```
fnoble: `assert(set_old_yet == 1);`?

fnoble:
Reviewed again, looks good, just a couple minor points to address.

fnoble:
looks great, maybe squash down some of the fix ups then I'd say its ready to go

kovach:
i'll test it again once hitl is up

src/amb_kf.c
fnoble: Ok, I think you are right about this now I'm reading it again

src/amb_kf.c
```diff
@@ -689,7 +696,9 @@ void rebase_mean_N(double *mean, const u8 num_sats, const u8 *old_prns, const u8
   u8 new_ref = new_prns[0];
```
fnoble: Early return if `old_ref == new_ref`?

src/amb_kf.c
```diff
@@ -715,12 +725,16 @@ static void assign_state_rebase_mtx(const u8 num_sats, const u8 *old_prns,
   u8 old_ref = old_prns[0];
   u8 new_ref = new_prns[0];
```
fnoble: Again, some kind of early return with a trivial rebase matrix?

fnoble:
Last todos:
 - Address remaining comments
 - Squash into single commit
 
@cbeighley : I looked at this fairly carefully and I think its correct, I don't think that in and of itself it could have caused the increased outliers but it could be changing the memory layout of the program and hence  changing the behavior of another bug. Also, the number of outliers hasn't changed much, just their distribution. I'd be inclined to merge it and continue tracking down the source of the remaining outliers.

cbeighley:
lgtm

cbeighley:
@kovach have we HITL tested this again? You said there were some minor changes you wanted to test, correct?

kovach:
I'm just going to merge it

PR #168, opened 2015-06-10
Added outlier detection and robustness to the KF and amb test

imh:
Because of cycle slips, false locks, and general mismodelling brought about by changing satellite sets, I opted not to use a data editing scheme, which would have allowed the possibility of locking up the filter forever and worked poorly in testing. Instead, I opted to use a more robust measurement update that essentially allows for heavier tails in the measurement model. If the measurement is good enough, it will be acted upon as normal, and the effect of this change is continuous and smooth. "Good" here refers to an adaptively scaled weighted sum of squares of KF innovations, weighted by their variances.
The KF also now returns whether it thought this data point was an outlier.

### Sacrifices made:
- We weight the SOS by their variances instead of by the whole covariance. We saved a few matrix multiplies and the solution of a system of equations.
- We are scaling by a slightly heuristic term, with connections to Huber's minimax estimators, but the connection is still kinda loose and ad hoc. It was simpler to implement and seems good enough.
- We implement the KF gain scaling in a slightly not-statistically-optimal way, but again, it seems good enough. The important thing here is that the covariances still update properly to match the mean.

### Some things to note:
- I got about 1 in 1000 false positives in early testing (SITL). They are mostly independent and therefore spaced out.
- I use an adaptive filter to decide the scale of what is and isn't an outlier, which dramatically decreases the incidence of long series of false positives.
- The timescale of adaptation is quite short, so this handles cycle slips/change points quite well, but won't handle long streams of outliers well. That said, there was an order of magnitude between the cutoff and the smallest outlier in my (small) training set. If we decrease the frequency of cycle slips, we can choose a longer timescale.
- When using a median filter rather than basic 1st (0th?) order IIR for the adaptation in the classifier, the median had a slightly lower false positive rate (FPR) for the same same FNR with less sensitivity between FNR and FPR. However, the median filter would have been more expensive and harder to implement.

### Plots/metrics:
In this plot, we are using the short timescale IIR with false positive rate a little over 1e-3, basically where it hooks over:
fpr versus safety

To connect a point on those curves to an actual cutoff, we consult this graph (this is for the short timeframe IIR, used in this PR):
fpr versus cutoff

cc @kovach, @fnoble, @mookerji

### DIFF
--- a/include/libswiftnav/amb_kf.h
+++ b/include/libswiftnav/amb_kf.h
@@ -22,6 +22,8 @@
 
 #define MAX_STATE_DIM (MAX_CHANNELS - 1)
 #define MAX_OBS_DIM (2 * MAX_CHANNELS - 5)
+#define KF_SOS_TIMESCALE 7.0f
+#define SOS_SWITCH 10.0f
 
 typedef struct {
   u32 state_dim;
@@ -34,9 +36,10 @@ typedef struct {
   double state_mean[MAX_STATE_DIM];
   double state_cov_U[MAX_STATE_DIM * MAX_STATE_DIM];
   double state_cov_D[MAX_STATE_DIM];
+  double l_sos_avg;
 } nkf_t;
 
-void nkf_update(nkf_t *kf, double *measurements);
+u8 nkf_update(nkf_t *kf, const double *measurements);
 
 void assign_phase_obs_null_basis(u8 num_dds, double *DE_mtx, double *q);
 void set_nkf(nkf_t *kf, double amb_drift_var, double phase_var, double code_var, double amb_init_var,
@@ -62,6 +65,9 @@ void rebase_covariance_udu(double *state_cov_U, double *state_cov_D, u8 num_sats
 
 void rebase_mean_N(double *mean, const u8 num_sats, const u8 *old_prns, const u8 *new_prns);
 void rebase_covariance_sigma(double *state_cov, const u8 num_sats, const u8 *old_prns, const u8 *new_prns);
+
+double get_sos_innov(const nkf_t *kf, const double *decor_obs);
+u8 outlier_check(nkf_t *kf, const double *decor_obs, double *k_scalar);
 
 #endif /* LIBSWIFTNAV_AMB_KF_H */
 
--- a/include/libswiftnav/amb_kf.h
+++ b/include/libswiftnav/amb_kf.h
@@ -31,22 +31,33 @@
 #define SOS_SWITCH 10.0f
 
 typedef struct {
+  /** The dimension of the state vector. */
   u32 state_dim;
+  /** The dimension of the observation vector. */
   u32 obs_dim;
+  /** The variance to use for the prediction update step (diffusion). */
   double amb_drift_var;
   /** The observation decorrelation matrix. Takes raw measurements and
    * decorrelates them. */
   double decor_mtx[MAX_OBS_DIM * MAX_OBS_DIM];
-  /* The observation matrix for decorrelated measurements. */
+  /** The observation matrix for decorrelated measurements. */
   double decor_obs_mtx[MAX_STATE_DIM * MAX_OBS_DIM];
-  /* The diagonal of the decorrelated observation covariance (for cholesky it's
+  /** The diagonal of the decorrelated observation covariance (for cholesky it's
    * ones). */
   double decor_obs_cov[MAX_OBS_DIM];
+  /** A basis for the left nullspace usual DGNSS observatio matrix, made of
+   * the DD line of sight vectors from the receivers to the sats. Used to
+   * project out the baseline's influence from the observations. */
   double null_basis_Q[(MAX_STATE_DIM - 3) * MAX_OBS_DIM];
+  /** The current state estimate. */
   double state_mean[MAX_STATE_DIM];
+  /** The upper unit triangular U matrix of the UDU decomposition of the
+   * covariance of the current state estimate. Stored dense. */
   double state_cov_U[MAX_STATE_DIM * MAX_STATE_DIM];
+  /** The diagonal D matrix of the UDU decomposition of the covariance of the current
+   * state estimate. Stored as a vector. */
   double state_cov_D[MAX_STATE_DIM];
-  /* A moving average of the log of the weighted sum of squares innovations. */
+  /** A moving average of the log of the weighted sum of squares innovations. */
   double l_sos_avg;
 } nkf_t;
 
--- a/include/libswiftnav/amb_kf.h
+++ b/include/libswiftnav/amb_kf.h
@@ -20,26 +20,39 @@
 #include "observation.h"
 #include "constants.h"
 
+/** \addtogroup amb_kf
+ * \{ */
+
 #define MAX_STATE_DIM (MAX_CHANNELS - 1)
 #define MAX_OBS_DIM (2 * MAX_CHANNELS - 5)
+/** The timescale for smoothing the innovation weighted sum of squares. */
 #define KF_SOS_TIMESCALE 7.0f
+/** The outlier cutoff for the highpassed innovation weighted sum of squares. */
 #define SOS_SWITCH 10.0f
 
 typedef struct {
   u32 state_dim;
   u32 obs_dim;
   double amb_drift_var;
-  double decor_mtx[MAX_OBS_DIM * MAX_OBS_DIM]; //the decorrelation matrix. takes raw measurements and decorrelates them
-  double decor_obs_mtx[MAX_STATE_DIM * MAX_OBS_DIM]; //the observation matrix for decorrelated measurements
-  double decor_obs_cov[MAX_OBS_DIM]; //the diagonal of the decorrelated observation covariance (for cholesky is ones)
+  /** The observation decorrelation matrix. Takes raw measurements and
+   * decorrelates them. */
+  double decor_mtx[MAX_OBS_DIM * MAX_OBS_DIM];
+  /* The observation matrix for decorrelated measurements. */
+  double decor_obs_mtx[MAX_STATE_DIM * MAX_OBS_DIM];
+  /* The diagonal of the decorrelated observation covariance (for cholesky it's
+   * ones). */
+  double decor_obs_cov[MAX_OBS_DIM];
   double null_basis_Q[(MAX_STATE_DIM - 3) * MAX_OBS_DIM];
   double state_mean[MAX_STATE_DIM];
   double state_cov_U[MAX_STATE_DIM * MAX_STATE_DIM];
   double state_cov_D[MAX_STATE_DIM];
+  /* A moving average of the log of the weighted sum of squares innovations. */
   double l_sos_avg;
 } nkf_t;
 
-u8 nkf_update(nkf_t *kf, const double *measurements);
+/** \} */
+
+bool nkf_update(nkf_t *kf, const double *measurements);
 
 void assign_phase_obs_null_basis(u8 num_dds, double *DE_mtx, double *q);
 void set_nkf(nkf_t *kf, double amb_drift_var, double phase_var, double code_var, double amb_init_var,
@@ -67,7 +80,13 @@ void rebase_mean_N(double *mean, const u8 num_sats, const u8 *old_prns, const u8
 void rebase_covariance_sigma(double *state_cov, const u8 num_sats, const u8 *old_prns, const u8 *new_prns);
 
 double get_sos_innov(const nkf_t *kf, const double *decor_obs);
-u8 outlier_check(nkf_t *kf, const double *decor_obs, double *k_scalar);
+double compute_innovation_terms(u32 state_dim, const double *h,
+                                double R, const double *U,
+                                const double *D, double *f, double *g);
+bool outlier_check(nkf_t *kf, const double *decor_obs, double *k_scalar);
+void update_kf_state(nkf_t *kf, double R, const double *f, const double *g,
+                   double alpha, double k_scalar,
+                   double innov);
 
 #endif /* LIBSWIFTNAV_AMB_KF_H */
 
--- a/src/amb_kf.c
+++ b/src/amb_kf.c
@@ -39,7 +39,7 @@
  * Preliminary integer ambiguity estimation with a Kalman Filter.
  * \{ */
 
-/**Calculation of vectors needed for the innovation scaling.
+/** Calculation of vectors needed for the innovation scaling.
  * We compute two vectors needed to make the Bierman update,
  * as well as the variance of the innovation.
  *
@@ -52,9 +52,9 @@
  * \param g         diag(D) * f.
  * \return          alpha = f * g + R, the innovation variance.
  */
-static double compute_innovation_terms(u32 state_dim, const double *h,
-                                       double R, const double *U,
-                                       const double *D, double *f, double *g)
+double compute_innovation_terms(u32 state_dim, const double *h,
+                                double R, const double *U,
+                                const double *D, double *f, double *g)
 {
   memcpy(f, h, state_dim * sizeof(double));
   /*  f = U^T * h. */
@@ -86,20 +86,25 @@ static double compute_innovation_terms(u32 state_dim, const double *h,
  * \param f         U^T * h
  * \param g         diag(D) * f
  * \param alpha     The innovation variance
- * \param k_scalar  A scalar to multiply the Kalman gain by (softens outliers)
+ * \param k_scalar  A scalar to multiply the Kalman gain by (softens outliers).
+ *                  Must be between 0 and 1 inclusive.
  * \param innov     The difference between the actual and predicted observation
  */
-static void update_kf_state(nkf_t *kf, double R, double *f, double *g,
-                            double alpha, double k_scalar,
-                            double innov)
+void update_kf_state(nkf_t *kf, double R, const double *f, const double *g,
+                     double alpha, double k_scalar,
+                     double innov)
 {
   DEBUG_ENTRY();
-
-  assert(k_scalar <= 1);
-  assert(k_scalar >= 0);
+  if (kf->state_dim == 0) {
+    return;
+  }
+  /* If we are scaling the update by 0, we aren't updating at all,
+   * so return early. */
   if (k_scalar == 0) {
     return;
   }
+  assert(k_scalar <= 1);
+  assert(k_scalar >= 0);
   u32 state_dim = kf->state_dim;
   double *U = kf->state_cov_U;
   double *D = kf->state_cov_D;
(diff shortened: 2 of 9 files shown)


include/libswiftnav/amb_kf.h
```diff
@@ -22,6 +22,8 @@
 
 #define MAX_STATE_DIM (MAX_CHANNELS - 1)
 #define MAX_OBS_DIM (2 * MAX_CHANNELS - 5)
+#define KF_SOS_TIMESCALE 7.0f
```
mookerji: Some comments describing what these constants are would be really helpful.

mookerji:
Looking at this now.

src/amb_kf.c
```diff
@@ -38,75 +43,102 @@
...
+ * \param R         The variance of the observation.
+ * \param U         KF state covariance U (not packed form)
+ * \param D         KF state covariance D (stored as a vector)
+ * \param f         U^T * h.
+ * \param g         diag(D) * f.
+ * \return          alpha = f * g + R, the innovation variance.
  */
-static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
-                                           double *U, double *D, double *k)
+static double compute_innovation_terms(u32 state_dim, const double *h,
```
mookerji: Unit test

src/amb_kf.c
```diff
@@ -162,21 +200,100 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
...
+ * but pretending they're independent, and then sum them.
+ * It's: sum_i (z - H * x)_i / (F * P * F' + R)_ii.
+ * If FPF'+R was actually diagonal and these measurements were independent,
+ * these would be chi square with kf->obs_dim degrees of freedom.
+ *
+ * \param kf        Kalman filter struct
+ * \param decor_obs Decorrelated observation vector
+ * \return          The weighted SOS.
+ */
+double get_sos_innov(const nkf_t *kf, const double *decor_obs)
```
mookerji: unit test?

tests/check_amb_kf.c
```diff
@@ -67,12 +68,73 @@ START_TEST(test_lsq)
 }
 END_TEST
 
+START_TEST(test_sos_innov)
```
mookerji: In terms of test coverage, I think it would be useful to have a filter test case using actual test data. Under ideal circumstances, we'd have a case with the KF from SITL where incorporating an outlier observation caused the original filter to fail.

src/amb_kf.c
```diff
@@ -38,75 +43,102 @@
...
+ *
+ * \param kf        The KF to update
+ * \param R         The measurement variance
+ * \param f         U^T * h
+ * \param g         diag(D) * f
+ * \param alpha     The innovation variance
+ * \param k_scalar  A scalar to multiply the Kalman gain by (softens outliers)
+ * \param innov     The difference between the actual and predicted observation
+ */
+static void update_kf_state(nkf_t *kf, double R, double *f, double *g,
```
mookerji: Unit test

src/amb_kf.c
```diff
@@ -139,7 +172,7 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
       k[i] += g[j] * U[i*state_dim + j]; /*  k = k + g[j] * U[:,j]. */
     }
     if (DEBUG) {
-      printf("gamma[%"PRIu32"] = %f\n", j, gamma[j]);
+      printf("gamma[%"PRIu32"] = %f\n", j, gamma);
```
mookerji: `log_info`?

src/amb_kf.c
```diff
@@ -38,75 +43,102 @@
...
+ * \param alpha     The innovation variance
+ * \param k_scalar  A scalar to multiply the Kalman gain by (softens outliers)
+ * \param innov     The difference between the actual and predicted observation
+ */
+static void update_kf_state(nkf_t *kf, double R, double *f, double *g,
+                            double alpha, double k_scalar,
+                            double innov)
+{
+  DEBUG_ENTRY();
+
```
mookerji: @fnoble what happens with `assert`s in the firmware. Are they noop's for the actual firmware artifact? In any case, wouldn't we want to have broader checking (like with the first check on `k_scalar`) with return values on errors?

tests/check_amb_kf.c
```diff
@@ -67,12 +68,73 @@ START_TEST(test_lsq)
...
+  kf.state_mean[0] = 1e20;
+  kf.state_mean[1] = 1e20;
+  /* Test with the predicted and actual observations wildly different. */
+  is_outlier = outlier_check(&kf, obs, &k_scalar);
+  /* This horrible of a match should always be an outlier,
+   * so k_scalar must be < 1. */
+  fail_unless(is_outlier);
+  fail_unless(k_scalar < 1);
+}
+END_TEST
```
mookerji: Are you testing for singular matrices?

src/amb_kf.c
```diff
@@ -38,75 +43,102 @@
...
   }
   else {
-    D_bar[0] = D[0] * R / gamma[0];
+    D_bar[0] = D[0] * R / gamma;
   }
   k[0] = g[0];
   U_bar[0] = 1;
   if (DEBUG) {
-    printf("gamma[0] = %f\n", gamma[0]);
+    printf("gamma[0] = %f\n", gamma);
```
mookerji: Similarly, see comment about logging macros below.

include/libswiftnav/amb_kf.h
```diff
@@ -34,9 +36,10 @@ typedef struct {
   double state_mean[MAX_STATE_DIM];
   double state_cov_U[MAX_STATE_DIM * MAX_STATE_DIM];
   double state_cov_D[MAX_STATE_DIM];
+  double l_sos_avg;
```
mookerji: nit: comment.

imh:
More explanation (for my own sake too) in hackpad under "Robust Estimation"

include/libswiftnav/amb_kf.h
fnoble: :+1: , you can use Doxygen style comments for them

include/libswiftnav/amb_kf.h
fnoble: Yeah, if you do that you should update the existing comments to Doxygen style

src/amb_kf.c
```diff
@@ -38,75 +39,96 @@
  * Preliminary integer ambiguity estimation with a Kalman Filter.
  * \{ */
 
-/** In place updating of the state cov and k vec using a scalar observation
- * This is from section 10.2.1 of Gibbs [1], with some extra logic for handling
- *    singular matrices, dictating that zeros from cov_D dominate.
+/**Calculation of vectors needed for the innovation scaling.
```
fnoble: nit: Space between `/**` and `Calculation`

src/amb_kf.c
fnoble: `assert`s work in the FW. They result in a `screaming_death()`. They should only be used of critical errors that would indicate a bug and have no sensible recovery.

src/amb_kf.c
```diff
@@ -38,75 +39,96 @@
...
+static void update_kf_state(nkf_t *kf, double R, double *f, double *g,
+                            double alpha, double k_scalar,
+                            double innov)
+{
+  DEBUG_ENTRY();
+
+  assert(k_scalar <= 1);
+  assert(k_scalar >= 0);
+  if (k_scalar == 0) {
+    return;
```
fnoble: Maybe a comment to explain why `k_scalar == 0` should just be an early return.

src/amb_kf.c
fnoble: Yes, `printf` is banned :)

src/amb_kf.c
fnoble: You know my preference is to get rid of these `DEBUG` sections cluttering the code :) Shouldn't be so necessary if the code its thoroughly unit tested.

src/amb_kf.c
```diff
@@ -162,21 +190,100 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
   DEBUG_EXIT();
 }
 
+/** Get the weighted sum of squared innovations.
+ * It's a normalized error metric. High is bad.
+ * More precisely, we square the difference between the predicted observations
+ * and the actual observations, divide them by their variances (FPF' + R),
+ * but pretending they're independent, and then sum them.
```
fnoble: Maybe clarify that the way we assume independence is by taking just the D from the UD factorization?

src/amb_kf.c
```diff
@@ -162,21 +190,100 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
...
+ * It's: sum_i (z - H * x)_i / (F * P * F' + R)_ii.
```
fnoble: Some of this should be Latexified for extra fancy but not a big deal

fnoble: But either way you should check the generated docs look readable.

src/amb_kf.c
fnoble: Oh wait, we are using U? But just assuming R is diagonal?

src/amb_kf.c
```diff
@@ -162,21 +190,100 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
...
+  matrix_multiply(kf->obs_dim, kf->state_dim, 1,
+                  kf->decor_obs_mtx, kf->state_mean, predicted_obs);
+  double hu[kf->obs_dim * kf->state_dim];
+  /* TODO use the fact that U is unit triangular to save a ton of time */
+  matrix_multiply(kf->obs_dim, kf->state_dim, kf->state_dim,
+                  kf->decor_obs_mtx, kf->state_cov_U, hu);
+  /* (H * U * D * U^T * H^T)_ii = (HU * D * HU^T)_ii
+   *                            = Sum_kl (HU_ik * D_kl * HU^T_li)
+   *                            = Sum_kl (HU_ik * D_kl * HU_il)
+   *                            = Sum_k (HU_ik * D_kk * HU_ik) */
```
fnoble: Seems like there is a bit of inconsistency in notation - we use F in some places and H in others? I think this comment should add in the step before where we go from `(F * P * F' + R)_ii` to `(H * U * D * U^T * H^T)_ii`.

src/amb_kf.c
```diff
@@ -162,21 +190,100 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
...
+  return sos;
+}
+
+/** Compute a scale factor to soften outliers, updating an outlier filter.
+ * We want to have some form of outlier detection that allows them
+ * (especially near the edge of the classifier) to influence the filter.
+ * This way, we soften the influence of any outliers, while eventually letting
+ * change-points through.
+ *
+ * I have found this moving average in log space to work best, only
```
fnoble: Maybe indicate that I == Ian Horn

src/amb_kf.c
```diff
@@ -162,21 +190,100 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
...
+  double l_sos = log(MAX(1e-10,sos));
+  double new_weight = 1.0f / KF_SOS_TIMESCALE;
+  *k_scalar =  MIN(1, SOS_SWITCH * exp(kf->l_sos_avg) / MAX(1e-10, sos));
+  kf->l_sos_avg = kf->l_sos_avg * (1 - new_weight) +
+                  l_sos * new_weight;
+  return (*k_scalar < 1);
+}
+
 /** In place updating of the state mean and covariances to use the (decorrelated) observations
- * This is directly from section 10.2.1 of Gibbs [1]
```
fnoble: long line is long

src/amb_kf.c
```diff
@@ -162,21 +190,100 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
...
+ * change-points through.
+ *
+ * I have found this moving average in log space to work best, only
+ * surpassed by a moving median. Both seem optimal with a timescale
+ * of 7 observations, as per detecting change points, but if we reduce
+ * the number of cycle slips, we may want to increase it.
+ *
+ * \param kf        The Kalman filter
+ * \param decor_obs Decorrelated observation vector
+ * \param k_scalar
```
fnoble: Undocumented

src/amb_kf.c
```diff
@@ -162,21 +190,100 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
...
+ * \return          A scalar to multiply Kalman gain by
```
fnoble: Looks like it actually returns a bool?

src/amb_kf.c
```diff
@@ -162,21 +190,100 @@ static void incorporate_scalar_measurement(u32 state_dim, double *h, double R,
...
+
 /** In place updating of the state mean and covariances to use the (decorrelated) observations
- * This is directly from section 10.2.1 of Gibbs [1]
+ * A modification of the update in section 10.2.1 of Gibbs [1].
+ * Changes are: -Scaling the kalman gain to deal with outliers.
+ *              -Some minor tweaks to handle singular matrices.
+ *
+ * \param kf        Kalman filter to be updated.
+ * \param decor_obs Decorrelated observation vector.
+ * \return          Whether we think the observations were bad.
```
fnoble: What are the specific return values?

src/amb_kf.c
```diff
@@ -208,18 +314,35 @@ static void make_residual_measurements(nkf_t *kf, double *measurements, double *
   }
 }
 
+/** The temporal update step of the KF.
```
fnoble: Maybe add a bit more high level explanation of what this is doing? Description and name don't really match and may be confused with `update_kf_state`. Explain that this is the update for the _covariance_ specifically.

src/amb_kf.c
```diff
@@ -208,18 +314,35 @@ static void make_residual_measurements(nkf_t *kf, double *measurements, double *
...
+ * To be really strict, since changes are equally likely on all channels,
+ * This should be += (I + 1 * 1^T)*var instead of I*var, but it's
+ * unlikely to be significant.
+ *
+ * \param kf The KF to be updated.
+ */
 static void diffuse_state(nkf_t *kf)
 {
+  double cov[kf->state_dim * kf->state_dim];
+  matrix_reconstruct_udu(kf->state_dim, kf->state_cov_U, kf->state_cov_D, cov);
```
fnoble: How come we now need to reconstruct?

src/amb_kf.c
```diff
@@ -231,17 +354,18 @@ void nkf_update(nkf_t *kf, double *measurements)
               kf->obs_dim, kf->decor_mtx, kf->obs_dim, /*  N, A, lda. */
               resid_measurements, 1); /*  X, incX. */
 
-  /*  predict_forward(kf);. */
+  /*  Temporal update */
```
fnoble: Temporal update seems like it could potentially be confusing as its non standard terminology and in our case only refers to an update of the covariance.

src/ambiguity_test.c
```diff
@@ -1520,7 +1528,7 @@ s8 determine_sats_addition(ambiguity_test_t *amb_test,
 u8 ambiguity_update_sats(ambiguity_test_t *amb_test, const u8 num_sdiffs,
                          const sdiff_t *sdiffs, const sats_management_t *float_sats,
                          const double *float_mean, const double *float_cov_U,
-                         const double *float_cov_D)
+                         const double *float_cov_D, u8 is_bad_measurement)
```
fnoble: I have some comments and questions about this function:
 - The control flow is very complex and intricate, can it be simplified?
 - The initial `sats_match` condition could be inverted and do an early return to reduce nested control flow
 - Couldn't we just to this at the top level in dgnss_management?

fnoble:
Done with first pass review. Some general comments:
 - Consider adding `asserts` to guard against NULL pointers and also to guard against invalid values that would indicate a bug, or other invariants you want to check (I see you did a bit of this but I'd like to emphasize this more in the future)
 - The unit tests so far test for normal operation of the functions. I think they would be a lot more useful if you consider all the cases where the function could fail - give it invalid inputs and check it responds correctly.

src/amb_kf.c
imh: Yeah, we use U. To get the diagonal elements of HPH'+R (which it should have been, per your other comment), we only need the diagonal elements of HPH' and the diagonal elements of R. R happens to be diagonal anyways, but for HPH', we need to use U. We aren't doing anything clever here with U and D.

src/amb_kf.c
imh: good catch. It should be written H throughout.

src/amb_kf.c
imh: We used to just update the KF cov by doing D += var. In some experimentation, the covariances are actually quite different doing it that way as opposed to doing UDU += var, so I switched it back.

src/ambiguity_test.c
imh: Yeah, we ought to clean it up. I made the basic inversion you mentioned and added some docs, but further refactoring is in order later.

tests/check_amb_kf.c
imh: In the case of this function, at least, it's just wrapping a bit of matrix algebra together. I think that would be overkill.

tests/check_amb_kf.c
imh: I am now!

src/amb_kf.c
imh: I'm not adding new prints, so lets revisit that in an unrelated PR

include/libswiftnav/amb_kf.h
imh: done.

src/amb_kf.c
imh: done

src/amb_kf.c
imh: I'm gonna stop just saying "done" now

src/amb_kf.c
fnoble: This still didn't get updated.

src/amb_kf.c
```diff
@@ -208,18 +340,38 @@ static void make_residual_measurements(nkf_t *kf, double *measurements, double *
...
-    kf->state_cov_D[i] += kf->amb_drift_var;
+    cov[i*kf->state_dim + i] += kf->amb_drift_var;
   }
+  matrix_udu(kf->state_dim, cov, kf->state_cov_U, kf->state_cov_D);
 }
 
-
-/** In place updating of the state mean and covariance. Modifies measurements.
+/** In place updating of the KF state mean and covariance.
+ * Does both the temporal and measurement updates to the KF.
```
fnoble: s/temporal/prediction/?

fnoble:
lgtm :shipit:

PR #200, opened 2015-07-21
Work saturdays

henryhallam:
I don't claim that this is the most rigorous possible way of handling time-of-week rollovers, but it's better than keeling over.

In this edition, the tracking channel doesn't attempt to keep track of the week number.  Instead when it's time to form a measurement, it uses gps_time_match_week() to set the measurement week number appropriately from the ephemeris, which may be equal, a week ahead or a week behind.

Bad results are still possible if we happen to have a cached ephemeris that's 7 days out of date and try to use it at just the wrong time.  A separate piksi_firmware update will regularly invalidate outdated ephemerides to prevent that from happening.

@mookerji @fnoble @cbeighley @denniszollo

### DIFF
--- a/include/libswiftnav/gpstime.h
+++ b/include/libswiftnav/gpstime.h
@@ -24,16 +24,20 @@
 
 /** Offset between GPS and UTC times in seconds.
  * Update when a new leap second is inserted and be careful about times in the
- * past when this offset was different. */
+ * past when this offset was different.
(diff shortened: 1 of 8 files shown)


cbeighley:
@henryhallam Can you add an issue for invalidating out of date ephemerides?

henryhallam:
Issue created https://github.com/swift-nav/piksi_firmware/issues/523

PR #202, opened 2015-07-24
Changed almanac functions to take const pointers.

ungureanuvladvictor:
Signed-off-by: Vlad Ungureanu <vvu@vdev.ro>

### DIFF
--- a/include/libswiftnav/almanac.h
+++ b/include/libswiftnav/almanac.h
@@ -39,11 +39,11 @@ typedef struct {
 
 /** \} */
 
-void calc_sat_state_almanac(almanac_t* alm, double t, s16 week,
+void calc_sat_state_almanac(const almanac_t* alm, double t, s16 week,
                             double pos[3], double vel[3]);
-void calc_sat_az_el_almanac(almanac_t* alm, double t, s16 week,
-                            double ref[3], double* az, double* el);
-double calc_sat_doppler_almanac(almanac_t* alm, double t, s16 week,
-                                double ref[3]);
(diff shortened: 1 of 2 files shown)


include/libswiftnav/almanac.h
```diff
@@ -39,9 +39,9 @@ typedef struct {
 
 /** \} */
 
-void calc_sat_state_almanac(almanac_t* alm, double t, s16 week,
+void calc_sat_state_almanac(const almanac_t* alm, double t, s16 week,
                             double pos[3], double vel[3]);
-void calc_sat_az_el_almanac(almanac_t* alm, double t, s16 week,
+void calc_sat_az_el_almanac(const almanac_t* alm, double t, s16 week,
                             double ref[3], double* az, double* el);
 double calc_sat_doppler_almanac(almanac_t* alm, double t, s16 week,
```
gsmcmullin: Why not make this alm const too?

gsmcmullin:
Looks good.  There are probably a lot of other missing const's around too.

henryhallam:
Good find. While you're at it, why not make the signatures of the functions fully const-correct? i.e. add `const` to `ref`.

mookerji:
+1 to this PR and Henry's suggestion as well.

mookerji:
While you're at it, fixing const correctness throughout the entire project wouldn't be terrible. Henry's already started some of this here: https://github.com/swift-nav/libswiftnav/pull/150/files

ungureanuvladvictor:
I will continue later on the const work. This PR will remain like this, I will do another one to refactor the remaining const things.

PR #206, opened 2015-07-29
Remove newlines from logs

mfine:
Used:

```
perl -pi -e 's/(log_.*)\\n"/\1"/' *.c
```

to strip the newlines out of the logging for most cases. Did the rest by hand. Left out the `printf`'s.

/cc @gsmcmullin


mfine:
Going to bring this in before it rots.

PR #215, opened 2015-08-28
SBAS integration

ungureanuvladvictor:
For the moment the tracking for SBAS is in bad shape (update_bit_sync needs to be rewritten) and observation module still assumes just L1 sats. Other than that these changes should make SBAS integration more easy.

This PR also has inside https://github.com/swift-nav/libswiftnav/commit/234b24511646350275be806071b48f2a67ce1707 which is in PR https://github.com/swift-nav/libswiftnav/pull/208 so they need to be applied in order.

\cc: @fnoble @kovach @gsmcmullin @denniszollo

### DIFF
--- a/include/libswiftnav/nav_msg.h
+++ b/include/libswiftnav/nav_msg.h
@@ -31,7 +31,7 @@
 #define DECISION_SIZE       (SBAS_NAVFLEN / 2 + 6)
 #define SBAS_BITS_UPDATE    ((SBAS_NAVFLEN + SBAS_NAVADDFLEN) / 2)
 #define SBAS_BITS_CHAINBACK (SBAS_NAVFLEN / 2)
-#define SBAS_DEC_SIZE       (SBAS_BITS_CHAINBACK / 8 + 1)
+#define SBAS_DEC_SIZE       (SBAS_BITS_CHAINBACK / 8)
 
 #define TOW_INVALID -1
 #define BITSYNC_UNSYNCED -1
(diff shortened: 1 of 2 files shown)


include/libswiftnav/ephemeris.h
```diff
@@ -25,18 +26,41 @@ typedef struct {
...
+  u8 valid;
+  u8 healthy;
+  signal_t sid;
+  u8 iod;
+  u16 toa;
+  u8 ura;
+  double a_gf0;
+  double a_gf1;
+} ephemeris_xyz_t;
+
```
denniszollo: Definitely need a comment here to explain how the ephemeris_t should be used.

denniszollo: I also thought we were planning to have an enum and a union of pointers for this.

include/libswiftnav/nav_msg.h
```diff
@@ -46,12 +61,48 @@ typedef struct {
   u8 bitsync_count;
   s32 bitsync_prev_corr[20];
   u32 bitsync_histogram[20];
+  u8 bit_length;
+} l1_legacy_nav_msg_t;
+
```
denniszollo: Put a reference to the section of the ICD where these terms are defined

src/almanac.c
```diff
@@ -36,8 +37,42 @@
...
  */
-void calc_sat_state_almanac(const almanac_t* alm, double t, s16 week,
-                            double pos[3], double vel[3])
+void sbas_calc_sat_state_almanac(const sbas_almanac_t* alm, double t,
+                                 double pos[3], double vel[3])
+{
+  u8 nweeks = t / 86400;
+  double tod = t - 86400 * nweeks;
+  double dt = tod - alm->t0;
+
```
denniszollo: explain these magic numbers (seconds in day and seconds in day/2)  explain this logic for determining the dt.

PR #218, opened 2015-09-18
check stuff

kovach:


kovach:
retest this please

DEVELOPMENT.md
```diff
@@ -0,0 +1,19 @@
+libswiftnav development
+=======================
+
+Follow these instructions if you intend to make changes to libswiftnav.
+
+Tools needed:
+ * doxygen
+ * pip install gcovr diff-cover
```
ungureanuvladvictor: it also needs convert from ImageMagick, without it cmake fails.

DEVELOPMENT.md
```diff
@@ -0,0 +1,19 @@
...
+To get started, run::
+  ./checks/setup-hooks.sh [DIR]
+
+from within the libswiftnav root directory. The default build-dir will be `build`.
+
+This makes the directory if it doesn't already exist, runs cmake with `Coverage` mode enabled
+inside it, and installs a git pre-commit hook that runs style and code-coverage checkers
+in this build directory. These checks require doxygen, gcovr, and diff-cover
+
+To manually run the coverage task, use `make coverage`. For syntax, use `make style`.
```
ungureanuvladvictor: make coverage worked, but make style reports:
`make: *** No rule to make target 'style'.  Stop.`

however make check-style works

checks/pre-commit
```diff
@@ -0,0 +1,7 @@
+set -e
+echo "precommit running. checking build located at ${subdir}."
+cd ${subdir}
+../checks/run-coverage.sh
```
mookerji: You should do some explicit error checking here on return codes.

mookerji: Also, run this through shellcheck: http://www.shellcheck.net/about.html

src/track.c
```diff
@@ -818,10 +824,10 @@ int nav_meas_cmp(const void *a, const void *b)
  * array at least `MIN(n_new, n_old)` long will ensure sufficient space.
  *
  * \param n_new Number of measurements in `m_new`
- * \oaram m_new Array of new navigation measurements
+ * \param m_new Array of new navigation measurements
```
mookerji: Ehhh, try to keep unrelated changes into a separate PR (advice for the future). This PR here is mostly a tooling change, right?

checks/pre-commit
scottswift: thanks, I'm checking it

DEVELOPMENT.md
scottswift: yes, an error in the readme

checks/pre-commit
scottswift: so this script basically just calls make now, which I think will print out enough information for the user to debug their problem - what would and explicit check on the code accomplish?

src/track.c
scottswift: this enables the tool to run without spitting out a bunch of errors

scottswift:
original checkpatch script taken from commit
9b9412dc7008f360c8e8ed10a654d3c8719f69d8
modifications by ungureanuvladvictor

checks/pre-commit
mookerji: I think I was looking at at run-coverage.sh and thought it would make sense to check if it fails and provide an error message if that happened. I also missed the "set -e". It's a simple script so I guess it doesn't matter.

PR #222, opened 2015-10-31
libswiftnav-python merge and packaging cleanup

mookerji:
libswiftnav's python bindings diverge form libswiftnav, at least in part because setup tools build them separately and against a global library installation, instead of a isolated, static library in Python's build directory. This has two big pieces:

1. Subtree merge of libswiftnav-python commits into libswiftnav.
2. Updates Python package conventions so that package can build without deprecated `distribute.py`. Does some setup so that setup tools can use the current version of libswiftnav in the repository.

(This is really just an offsite to work on libswiftnav-python.)

/cc @mfine @fnoble @benjamin0 @kovach


mookerji:
ok to test

mookerji:
@scottswift says :+1:

PR #225, opened 2015-11-10
Dgnss refactor

scottswift:
update to PR #150


scottswift:
bench testing caused hard-faults. probably need to redo this again given how much time has passed.

PR #226, opened 2015-11-12
libswiftnav cython updates

mookerji:
Updated Cython bindings for libswiftnav.

(Also known as the "A bird in the hand is worth two in the bush" commit).

WIP! Now seems as good a time as any to get some early feedback. I'm currently porting some of the regular libswiftnav tests to verify that some things work. Note that some of the commits here are from the merging of libswiftnav-python with libswiftnav (https://github.com/swift-nav/libswiftnav/pull/222), which hasn't been merged yet. 

The important commit in question is this one: https://github.com/swift-nav/libswiftnav/commit/9fb344df4a450d7b93d8d629f2edecbbccc8e9e2 

Alas, there is unfortunately still a fair bit of work left to do....

A massive monolithic commit (sorry!) containing:
- Updated coverage on bindings
- Removes c-specific Cython declaration files and other detritus
- Very mild unit test coverage on ported structs and functions

Some notes:
- Documentation is still lacking, but perhaps we can just point people at the corresponding C documentation or something.
- Some of the changes here are trying to minimize boiler plate, such as struct initialization for CPython wrapper objects. Many of the Cython objects are missing boilerplate setters and getters, but hopefully that can be handled some more nicer way.
- Calling conventions are a bit inconsistent in the bindings and the underlying C library. See, I & Q quadratures vs complex p in the tracking loop stuff: track.c.

/cc @mfine @fnoble @kovach @benjamin0 @JoshuaGross @henryhallam @gsmcmullin @jacobmcnamee


mookerji:
ok to test

PR #228, opened 2015-11-15
Updated Cython bindings for libswiftnav.

mookerji:
Updated Cython bindings for libswiftnav.

(Also known as the "A bird in the hand is worth two in the bush" commit).

WIP! Now seems as good a time as any to get some early feedback. I'm currently porting some of the regular libswiftnav tests to verify that some things work. 

Alas, there is unfortunately still a fair bit of work left to do....

A massive monolithic commit (sorry!) containing:
- Updated coverage on bindings
- Removes c-specific Cython declaration files and other detritus
- Very mild unit test coverage on ported structs and functions

Some notes:
- Documentation is still lacking, but perhaps we can just point people at the corresponding C documentation or something.
- Some of the changes here are trying to minimize boiler plate, such as struct initialization for CPython wrapper objects. Many of the Cython objects are missing boilerplate setters and getters, but hopefully that can be handled some more nicer way.
- Calling conventions are a bit inconsistent in the bindings and the underlying C library. See, I & Q quadratures vs complex p in the tracking loop stuff: track.c.

/cc @mfine @fnoble @kovach @benjamin0 @JoshuaGross @henryhallam @gsmcmullin @jacobmcnamee @akleeman


python/swiftnav/almanac.pyx
```diff
@@ -117,23 +78,13 @@ cdef class Almanac:
...
-      week = -1
-
-    almanac_c.calc_sat_az_el_almanac(&self.almanac, t, week,
-                                     &ref_[0], &az, &el)
-
+    calc_sat_az_el_almanac(&self._thisptr, t, week, &ref_[0], &az, &el)
     return (az, el)
 
-  def calc_doppler(self, t, ref, week=None):
+  def calc_doppler(self, t, ref, week=-1):
```
akleeman: I think the docs need to be updated below regarding the default week value.

python/swiftnav/amb_kf.pyx
```diff
@@ -7,40 +7,29 @@
...
-  memcpy(&state_mean_[0], kf.kf.state_mean, kf.state_dim * sizeof(double))
-  memcpy(&state_cov_U_[0,0], kf.kf.state_cov_U, kf.state_dim * kf.state_dim * sizeof(double))
-  memcpy(&state_cov_D_[0], kf.kf.state_cov_D, kf.state_dim * sizeof(double))
-
-  return state_mean_, UDU_decomposition(state_cov_U_, state_cov_D_)
-
+from observation cimport *
+import numpy as np
 
+# def filter_update(KalmanFilter kf, obs):
```
akleeman: should these comments be removed?

python/swiftnav/memory_pool.pxd
```diff
@@ -0,0 +1,60 @@
+# Copyright (C) 2015 Swift Navigation Inc.
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+from common cimport *
```
akleeman: Not sure about standard practice in cython, but in python 'import *' is best avoided.

python/swiftnav/pvt.pyx
```diff
@@ -7,73 +7,53 @@
...
-  s = Solution()
-  pvt_c.calc_PVT(n_used, nav_meas_array, False, &(s.soln), &(s.dops))
-
-  free(nav_meas_array)
-
-  return s
+    nav_meas_[n] = (<NavigationMeasurement?>nav_meas[n])._thisptr
+  dops_ = DOPS()
+  soln = GNSSSolution()
+  cdef u8 ret = calc_PVT(n_used, nav_meas_, disable_raim, &soln._thisptr, &dops_._thisptr)
```
akleeman: I noticed that when I provided invalid inputs to calc_PVT (before this change) it would sometimes fail, but still return solutions with x, y, z values of 0.  Perhaps this would be a good place to check for success and raise and propagate any errors if they occurred?

python/swiftnav/almanac.pxd
```diff
@@ -7,7 +7,27 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
```
akleeman: I find it helpful to have short descriptions about what a module contains, not sure if something similar is found elsewhere in the supporting code for this, but it could be beneficial in these modules as well.

akleeman:
I'm not familiar enough with the previous code (or cython) to give full approval, but looks good to me.

Aside: Is this code timing critical?  If so, are there benchmarks?  I know, particularly with cython, minor changes can have significant impacts, for better or worse, on run time.  Then again, if this is all for analysis / out of chip computation that may be overkill.

python/swiftnav/almanac.pxd
mookerji: Yeah, we need to port some namespace documentation. I don't know how much of a slog it'll be to transition over docstrings, but we should probably do that for commonly used functions.

mookerji:
There are no benchmarks. there's a lot of extra random memory allocation and copying where there doesn't need to be (see any place where an array of CPython objects gets mapped to structs and vice versa). I'm going to probably refactor those places so we can get by more efficiently once more tests are in place.

python/swiftnav/pvt.pyx
mookerji: Interesting. yeah, there's a lot of checking to be done.

python/swiftnav/memory_pool.pxd
mookerji: It's likely the same...

python/swiftnav/almanac.pyx
mookerji: Yep!

python/swiftnav/amb_kf.pyx
mookerji: A bunch of them need to be removed.

PR #232, opened 2015-11-17
Add SBAS PRN codes.

gsmcmullin:
/cc @jacobmcnamee


src/prns.c
```diff
@@ -21,17 +24,30 @@ static const u8 ca_codes[32][128];
...
  *
- * \param prn PRN number.
+ * \param sid Signal ID.
  *
  * \return A pointer to an array containing the C/A code PRN corresponding to
  *         the PRN number <code>prn</code>, packed one bit per chip.
  */
-const u8* ca_code(u8 prn) {
-  return ca_codes[prn];
+const u8* ca_code(gnss_signal_t sid)
```
gsmcmullin: How do we extend this to other bands?  Do we need to?
@fnoble

include/libswiftnav/signal.h
```diff
@@ -15,14 +15,9 @@
...
-#define GAGAN_SATS 2
-#define MSAS_SATS 2
-#define SDCM_SATS 2
-#define SBAS_SATS (WAAS_SATS + EGNOS_SATS + GAGAN_SATS + MSAS_SATS + SDCM_SATS)
-#define ALL_SATS (GPS_L1_SATS + SBAS_SATS)
+#define NUM_SATS_GPS 32
+#define NUM_SATS_SBAS 22
+#define NUM_SATS (NUM_SATS_GPS + NUM_SATS_SBAS)
 
 #define CONSTELLATION_GPS 0
```
jacobmcnamee: Can we make constellation and band enums with CONSTELLATION_COUNT and BAND_COUNT at the bottom?

PR #233, opened 2015-11-18
Pseudorange calculation improvements

fnoble:
**What:** Currently we choose a reference satellite and set its pseudorange arbitrarily to a nominal pseudorange value. This doesn't affect the PVT solution but does have a few consequences, mainly that our observations are not compatible with a lot of standard GPS software, and more importantly the pseudorange rate will no longer reflect the rate of change of the geometric range. This makes it impossible to do carrier smoothed code and other techniques where you are combining code and carrier measurements.

**How:** This PR adds the ability to specify the GPS time corresponding to the time of reception. This makes the pseudorange simple to calculate and close to the true range. piksi_firmware has a function to translate the receiver time to GPS time once we have done at least one position solution, so after one solution we can begin to use the GPS time of reception.

Also cleans up this function and add doxygen documentation.

**Testing:** Bench test looks fine, will HITL test.

/cc @benjamin0 @gsmcmullin @sashamitelman

### DIFF
--- a/include/libswiftnav/track.h
+++ b/include/libswiftnav/track.h
@@ -233,10 +233,12 @@ float cn0_est(cn0_est_state_t *s, float I, float Q);
 
 void calc_navigation_measurement(u8 n_channels, channel_measurement_t meas[],
                                  navigation_measurement_t nav_meas[],
-                                 double nav_time, ephemeris_t ephemerides[]);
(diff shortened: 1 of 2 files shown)


ljbade:
Closing in favour of https://github.com/swift-nav/libswiftnav/pull/310

PR #231, opened 2015-11-17
Generalisation of ephemeris for SBAS support.

gsmcmullin:
/cc @jacobmcnamee

PR #235, opened 2015-11-19
remove indexing by PRN

jacobmcnamee:
API changes to 
make_propagated_sdiffs()
calc_navigation_measurement()
check_lock_counters()

all of which now take in ephemeris pointers corresponding to the signals in other arrays instead of the entire global ephemeris array

### DIFF
--- a/include/libswiftnav/observation.h
+++ b/include/libswiftnav/observation.h
@@ -48,14 +48,14 @@ u8 make_propagated_sdiffs_wip(u8 n_local, navigation_measurement_t *m_local,
 u8 make_propagated_sdiffs(u8 n_local, navigation_measurement_t *m_local,
                           u8 n_remote, navigation_measurement_t *m_remote,
                           double *remote_dists, double remote_pos_ecef[3],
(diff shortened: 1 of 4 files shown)


gsmcmullin:
The const correctness of `ephemeris_good` is fixed in #231.  This can be rebased on that to fix the build problem.
Otherwise, lgtm.

jacobmcnamee:
retest this please

gsmcmullin:
add to whitelist

jacobmcnamee:
retest this please

gsmcmullin:
ok to test

fnoble:
Sorry I didn't get to it in time but there was a reason why we had `calc_navigation_measurement_()` taking arrays of pointers instead of flat arrays - its used by peregrine. Would be great if you could make a PR to add it back in.

gsmcmullin:
`calc_navigation_measurement` now takes an array of pointers, so it can be used by Peregrine instead.  Piksi firmware and Peregrine will now use the same interface.

PR #240, opened 2015-12-01
Issue warnings if calc_PVT fails.

akleeman:
Previously zeros were returned.

This sort of warning logic should probably be added to other parts of the python/c interface as well.

### DIFF
--- a/python/swiftnav/pvt.pyx
+++ b/python/swiftnav/pvt.pyx
@@ -7,6 +7,8 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
+import warnings
+
 cimport pvt_c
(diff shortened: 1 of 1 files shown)


python/swiftnav/pvt.pyx
```diff
@@ -72,8 +84,11 @@ def calc_PVT(nav_meas):
     nav_meas_array[n] = (<NavigationMeasurement?>nav_meas[n]).meas
 
   s = Solution()
-  pvt_c.calc_PVT(n_used, nav_meas_array, False, &(s.soln), &(s.dops))
-
+  cdef s8 ret
```
mookerji: Looks good. Note that you can just do `cdef s8 ret = ...`.

mookerji:
Hmmm, so it was returning 0's for the PVT solution (ECEF) but not returning a problem? what error code was there before.

akleeman:
@mookerji As far as I can tell there isn't any error logic.  If the routine fails, the c code will fill the Solution object with zeros ... but the wrapper just passes that on.  Certainly no information about why the routine failed is propagated.

PR #242, opened 2015-12-04
Generalize bit sync, separate from nav message decoding

jacobmcnamee:


python/swiftnav/nav_msg.pyx
```diff
@@ -16,8 +16,8 @@ from signal cimport gnss_signal_t
 
 cdef class NavMsg:
 
-  def __cinit__(self):
-    nav_msg_init(&self._thisptr)
+  def __cinit__(self, sid):
```
mookerji: Yeah, so this should be something like this...

```python
def __cinit__(self, GNSSSignal sid):
    nav_msg_init(&self._thisptr, sid._thisptr)
```

since you have to pass the pointer to the method. You should have gotten a typechecking error here on Travis probably.

(Also, thanks for updating the bindings!)

python/swiftnav/nav_msg.pxd
```diff
@@ -10,6 +10,7 @@
 from common cimport *
 from ephemeris cimport *
 from libcpp cimport bool
+from signal cimport *
 
 cdef extern from "libswiftnav/nav_msg.h":
   enum:
```
mookerji: I think you also need to add the new struct fields to the extern definition.

python/swiftnav/nav_msg.pyx
jacobmcnamee: Actually SIDs are passed by value for better or worse.

python/swiftnav/nav_msg.pxd
jacobmcnamee: Updated now.

python/swiftnav/nav_msg.pxd
```diff
@@ -21,21 +22,22 @@ cdef extern from "libswiftnav/nav_msg.h":
     BIT_POLARITY_UNKNOWN
 
   ctypedef struct nav_msg_t:
```
mookerji: Yeah, you also need to declare: `BIT_LENGTH_MAX`. the semicolons are also redundant.

python/swiftnav/nav_msg.pyx
mookerji: err, yes. so `_thisptr` actually refers to the underlying C struct, it's not actually a pointer, so `sid._thisptr` actually passes it by value.

include/libswiftnav/bit_sync.h
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef LIBSWIFTNAV_BIT_SYNC_H
+#define LIBSWIFTNAV_BIT_SYNC_H
+
+#include "common.h"
```
ljbade: This needs to be changed to new include style `<libswiftnav/common.h>`

include/libswiftnav/bit_sync.h
```diff
@@ -0,0 +1,47 @@
...
+#include "signal.h"
```
ljbade: Same ^

ljbade:
@jacobmcnamee looks good to me, except for fixing the includes per comments ^

PR #243, opened 2015-12-04
fix missing factor of 2 in frequency descriminators

jacobmcnamee:


ljbade:
@jacobmcnamee Is this ready for review?

fnoble:
@jacobmcnamee why do you think this is necessary? can you provide a reference for the updated math?

PR #253, opened 2015-12-15
Add support for QZSS (WIP - do not merge)

ljbade:
Currently working on adding support for the Japanese QZSS constellation.

It is designed to be compatible with GPS and only a few tweaks will be needed to support this system.


fnoble:
add to whitelist

Issue #159, opened 2015-06-07
Consistency-check IODC/IODE across all three subframes of nav msg


henryhallam:
/cc @fnoble @mookerji 
Per the interface spec, the first three subframes (which comprise the clock correction curve and the ephemeris) each include an Issue Of Data field - IODC in subframe 1 and IODE in each of subframes 2 and 3.  This is provided so the receiver can ensure that it uses data from a complete set, and doesn't mix and match across ephemeris updates.  Presently, we load IODE into the ephemeris struct and never look at it, and we don't even record IODC at all.

This isn't an urgent problem because of the rather conservative logic in nav_msg.c:process_subframe(), which will only call decode_ephemeris() when it has received 3 valid (i.e. passing parity) subframes in a row, in the strict order 1,2,3.  I'm pretty sure the changeover of the actual transmitted message from one set to the next occurs on a 30 second boundary, i.e. between transmission of subframe 5 and 1, so I don't think it's possible to do any of that kind of mixing + matching with the current code.  But by using the IODC/IODE as intended, we'd be able to improve time to first single-ended fix by up to 12 seconds in some cases.  And it's a useful belt-and-braces check.

fnoble:
I believe the ICD says its ok if IODC != IODE?

henryhallam:
"The issue of ephemeris data (IODE) term shall provide the user with a convenient means for
detecting any change in the ephemeris representation parameters. The IODE is provided in both
subframes 2 and 3 for the purpose of comparison with the 8 LSBs of the IODC term in subframe
1. Whenever these three terms do not match, a data set cutover has occurred and new data must
be collected. The timing of the IODE and constraints on the IODC and IODE are defined in
paragraph 20.3.4.4."

ljbade:
Done in https://github.com/swift-nav/libswiftnav/pull/255

PR #255, opened 2015-12-16
More careful handling of GPS NAV data

### LINKED ISSUE #254: Improvements needed to handling of GPS nav data
There are a few improvements needed to the current decoding of GPS NAV data to correctly follow the GPS interface spec and to detect bad satellites. I first noticed these last year, and this code is still the same.

Main issues:
- alert flag of HOW word is not checked
- IODC and IODE x 2 are not compared - see https://github.com/swift-nav/libswiftnav/issues/159
- URA is not even extracted and should be checked for a value of 15, indicating a bad satellite
- the curve fit valid interval is not correctly calculated, currently hard coded to 4 hours
- health flags not correctly interpreted

ljbade:
Fixes #254 and #159.

So far I have implemented:
- IODE and IODC cross checks. Needs testing for when they don't match (e.g. cut over)
- URA extraction and checking for 15 unhealthy value, also works out the value in meters and stores in ephemeris struct. Could be used for weighting in the position solution. Needs testing, so far I have only seen satellites with 2.0 or 2.8m URA.
- all 6 bits of health flags are correctly checked. Currently any unhealthy signal will mark entire satellite as unhealthy in the stored ephemeris. Could be made smarter to only check for unhealthy L1 C/A.
- curve fit interval is correctly calculated and used. Needs testing via simulator for cases when GPS sats go into extended operation mode (e.g. major catastrophe takes out both GPS control centres)

Questions I have:
- we should use IODE instead of IODC for comparing ephemeris (since IODC is unique within a week, IODE only 6 hours) - what code needs to be changed here?
- I have added some new members to the ephemeris structs, does the binary interface need updating?
- how can I modify `nav_msg.c` to extract and store per sat the alert flag, separate from ephemeris struct (since this could change between ephemeris updates)?
- how can I modify the position calculation to ignore sats with an alert or unhealthy bits

@cbeighley Who is best to help me with these?


ljbade:
One of the python tests is failing, and I'm not sure why.

ljbade:
I wonder if the alert flag should also be part of the observation messages?

mookerji:
Hey @ljbade, 

- The ephemeris test breaking is a unit test of the libswiftnav bindings on the Ephemeris object. If the semantics of the `ephemeris_good` function have changed, then the test should be updated accordingly.
- By "binary interface" do you mean the fields of the libsbp ephemeris message? If so, the answer is largely Yes. The rover and the base station exchange ephemeris messages over the radio, and we also consume them externally off the device. If those fields are only used internally, then don't bother. 
- For the position solution stuff, I recommend reading [calc_PVT](https://github.com/swift-nav/libswiftnav/blob/master/src/pvt.c) and adding an exclusion there. We currently handle exclusions for RAIM, so that might be a good example to crib off of.

ljbade:
@mookerji For the alert flag, I did find this bit of code: https://github.com/swift-nav/piksi_firmware/commit/9d0e7bb66a96a2ef822899082608384f2d9d0c5c#diff-3db690123986aa7cf0d7d32c546cebe0R498

I extended the exclusion of unhealthy ephemeris sats to include the alert flag...

As for `libsbp` I have added to the ephemeris.h `iodc` `ura` and `fit_interval`. There was already a field in `libsbp` for IODC (and the python bindings) but it was not filled out...

So I think I only need `fit_interval` at this stage. `ura` would be needed if we use that down the road in the position calculation to give lower priority to less accurate sats.

mookerji:
@ljbade , well, if you need to update libsbp, I updated our versioning policy here: https://github.com/swift-nav/libsbp/pull/322. 

The specification for stable messages is fixed. You'll need to add a new SBP message type for an ephemeris with additional fields, and regenerate the C, Python, and Haskell bindings (our most frequently used bindings). Afterwards, propagate them to the firmware. My recommendation is to add all the remaining fields (`fit_interval` and `ura`) to the ephemeris message if you intend on using them in the future. Adding new messages causes a bit of churn in our tools, so best to do that only once.

That being said, the last time we audited the message definitions, we either overlooked the fields you mentioned (from the IDC?) or decided not to include them. I don't remember which one. @fnoble @henryhallam ?

mfine:
Right, the ephemeris structures between libswiftnav and libsbp are getting mapped in [sbp_utils.c#L267-L329]](https://github.com/swift-nav/piksi_firmware/blob/master/src/sbp_utils.c#L267-L329). If the SBP ephermeris gets extended, it'll need a new msg and a rev of the library for compatibility. Pretty straightforward - the ephemeris has been extended a number of times already (see  the already deprecated ephemeris messages).

ljbade:
@mfine Would it be worth waiting for a review to see what is needed to handle the recently added SBAS ephemeris before creating a new message definition?

I think we can merge this PR and come back to update `libsbp` in a later PR.

ljbade:
I have changed the ephemeris comparison functions to use IODC (as well as IODE).

I have also added checks for the alert flag to the relevant firmware branch.

Next thing is to work on the test cases.

ljbade:
I have updated the `ephemeris_equal` to handle the new GPS and SBAS fields.

I have also updated `test_ephemeris` to take into account changes around `fit_interval`.

I also want to implement a few more in there I created TODOs for. But my Python is rusty.

Can I update the `eph` object to change its fields directly, or do I have to create a whole new object?

ljbade:
The python test is still failing on `assert eph.good(g.GpsTime(**{ 'wn': 1867, 'tow': 518400.0,}))`

@mookerji Are you able to take a closer look at why this is still failing?

mfine:
@ljbade that sounds good - bundling up the changes around a message format update makes a lot of sense - it's not a painfree process.

mookerji:
@ljbade yeah, I can check it out in a bit.

python/tests/test_ephemeris.py
```diff
@@ -44,6 +46,12 @@ def test_sat_state():
                                   'af1': 3.637978807091713e-12,
                                   'af2': 0.0}})
   assert eph.good(g.GpsTime(**{ 'wn': 1867, 'tow': 518400.0,}))
+  eph.healthy = 0
```
mookerji: Ack! Figured out what was going on here. 

checkout https://github.com/swift-nav/libswiftnav/blob/tidy-gps-nav/python/swiftnav/ephemeris.pyx#L28.

You'll need to add `fit_interval` to be initialized. Currently, it gets set to the initialized value, which is 0, so this test fails.

Lastly, you can't set properties for the Ephemeris object, since the property setter isn't defined. Consequently, this next line at 49 will fail. You could inline another Ephemeris object with `healthy` set to 0, or you can define property setters for those fields.

mookerji: I can't believe I'd say this but maybe we should bite the bullet and migrate to using Swig.

ljbade:
retest this please

ljbade:
This should be ready for review.

ljbade:
retest this please

ljbade:
@jacobmcnamee or @gsmcmullin Can you review this please?

src/ephemeris.c
```diff
@@ -257,7 +254,65 @@ u8 ephemeris_good(const ephemeris_t *eph, gps_time_t t)
...
+}
+
+/** Convert a GPS URA index into a value.
+*
+* |param index URA index.
+* |return the URA in meters.
+*/
+float decode_ura_index(const u8 index) {
+  static float values[16] = {
+    2.0f,    // 0
```
gsmcmullin: Consider using c99 style: `[0] = 2.0f,`, or just removing the comments showing the array indices.

gsmcmullin:
This looks great.
Also, thanks for adding some tests for the xyz ephemeris.

src/ephemeris.c
ljbade: Ah yeah I will do it like that.

ljbade:
@gsmcmullin I made the change to the URA table.

Let me know if you now happy to merge.

gsmcmullin:
Go ahead and merge.

ljbade:
Hmm I just noticed that I overlooked table Table 20-XII for calculation of URA. Making note for me to come and amend that.

PR #257, opened 2015-12-18
Add alert and healthy flags to solution structs

ljbade:
Continuation of #255. Needed so in https://github.com/swift-nav/piksi_firmware/pull/588 I can exclude unhealthy sats from the solution.

@gsmcmullin Can you review?

### DIFF
--- a/include/libswiftnav/constants.h
+++ b/include/libswiftnav/constants.h
@@ -34,7 +34,7 @@
 /** The official GPS value of Pi.
  * This is the value used by the CS to curve fit ephemeris parameters and
  * should be used in all ephemeris calculations. */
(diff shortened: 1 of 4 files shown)


ljbade:
Before merge I will add a fix since I just noticed that I overlooked table Table 20-XII for calculation of URA.

ljbade:
I should also include the fixed GPS_PI

ljbade:
@gsmcmullin Ready for PR again after fixing last bits.

ljbade:
Going to merge this since there was no review comments and this is only a small change. Functionality will change in the corresponding firmware PR that will need more thorough testing.

PR #259, opened 2015-12-19
Fixes to lesq_solve_raim bindings and tests

fnoble:
When RAIM repair is performed we were returning a residuals array that was too long with junk elements at the end.

cc: @mookerji @ljbade


ljbade:
Looks good.

mookerji:
lgtm!

PR #262, opened 2015-12-21
Refactor #includes to avoid namespace conflict

ljbade:
In the process of working on UTC functions @fnoble and I decided to rename `gpstime.h` to `time.h` since it will have both GPS and UTC time functions.

However this caused a compile error due to conflict with the standard C library `time.h`.

We already put includes in their own name space via `include/libswiftnav` but all our code was compiled with the headers relative to `include/libswiftnav` not `include/`.

This PR changes everything to correctly use the `include/` path. This results in includes like `#include <libswiftnav/common.h>`.

I took the opportunity to make every file's `#include` lines have consistent formatting and ordering.

Finally this PR also renames `gpstime.h` to `time.h` in preparation for the UTC functions.

This patch will likely break other PRs until they get rebased of this.

@fnoble You mind reviewing?


ljbade:
Can someone from @swift-nav/firmware review this?

ljbade:
@fnoble Said I should merge.

PR #263, opened 2015-12-21
Update README.md to show build status for master.

akleeman:
The build status was showing the most recent libswiftnav build (including PRs), so that while looking at the master branch you may see a red "build failed" which actually corresponded to a development branch.

https://github.com/letsencrypt/letsencrypt/issues/927

This change makes it so the build status (on all branches) shows an indicator of the master build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 libswiftnav
 ===========
 
-[![Build status][1]][2]
+[![Build Status](https://travis-ci.org/swift-nav/libswiftnav.svg?branch=master)](https://travis-ci.org/swift-nav/libswiftnav)
 
 Libswiftnav is a platform independent library that implements GNSS functions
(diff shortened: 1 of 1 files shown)


mookerji:
Why not update the reference for `[1]` at the bottom of the README?

mookerji:
(also, good catch.)

akleeman:
Good call, didn't think of modifying the references.

README.md
```diff
@@ -1,7 +1,7 @@
 libswiftnav
 ===========
 
-[![Build status][1]][2]
+libswfitnav/master: [![Build status][1]][2]
```
mookerji: Typo!

You could also have have `master`, since `libswiftnav` is redundant. 

Feel free to merge after that.

Issue #265, opened 2015-12-22
Ideas on GPS week rollover


ljbade:
The next GPS rollover is in 2019, which isn't that far away and likely in the lifetime of current receivers.

The cause is that GPS WN is broadcast to 10 bits so is mod 1024 of the real WN, which is currently between 1024-2047 since the last rollover in 1999.

Currently we hardcode that we are in 1024-2047, so in WN 2048 the receiver will warp back to 1999. Many of our time functions will also break as they do not check for roll over.

At a minimum the current WN range should be a setting to allow users to update it in the future.

We should also encode the compile date of the firmware to allow us to compute a valid WN for 20 years after shipping.

Finally we could have the Piksi console update these settings from the local computer's clock, so that a Piksi will have a valid date for 20 years after connecting it to the console.

For an example of how u-blox handle this, read https://forum.u-blox.com/index.php?qa=60&qa_1=what-week-rollover-roll-over-and-handled-blox-gnss-receivers

/cc @mookerji @fnoble

ljbade:
A similar technique is mentioned in http://gauss.gge.unb.ca/gpsworld/gpsworld.november98.pdf

They also point out a real time clock is useful for this too.

PR #267, opened 2015-12-22
Add gps_adjust_week_cycle for handling week rollover

ljbade:
First step at addressing #265 

Next step will be to make the reference week a firmware setting,.

### DIFF
--- a/src/time.c
+++ b/src/time.c
@@ -98,4 +98,21 @@ void gps_time_match_weeks(gps_time_t *t, const gps_time_t *ref)
     t->wn++;
 }
 
+/** Adjust the week number of t to correctly refelct the current week cycle.
+ *
+ * Assumes the current week number cannot be earlier than the reference WN. So
+ * will return the correct WN for at most 20 years after the reference WN.
+ *
+ * \param wn_raw Raw week number from NAV data stream that is modulo 1024
+ * \param wn_ref Reference week number that is from some point in the past
+ * |return The abosolute week number counted from 1980
+ */
+u16 gps_adjust_week_cycle(u16 wn_raw, u16 wn_ref)
+{
+  if (wn_raw >= wn_ref)
+  return wn_raw;
+
+  return wn_raw + 1024 * ((wn_ref + 1023 - wn_raw) / 1024);
+}
+
 /** \} */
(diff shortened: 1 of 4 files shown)


ljbade:
@fnoble I recall you had concern about hard coding `GPS_WEEK_REFERENCE`?

This is my reasoning for having it in libswiftnav:
- For a libswiftnav unit tests I needed this in here
- In the firmware this value will only be used as the compile time default
- The user will be able to change this via a setting

Until I figure out how to calculate current GPS WN at compile time, this number will have to be updated by hand for a new firmware release.

ljbade:
@swift-nav/firmware Ready for review

src/time.c
```diff
@@ -101,3 +101,21 @@ void gps_time_match_weeks(gps_time_t *t, const gps_time_t *ref)
...
+
+/** Adjust the week number of t to correctly refelct the current week cycle.
+ *
+ * Assumes the current week number cannot be earlier than the reference WN. So
+ * will return the correct WN for at most 20 years after the reference WN.
+ *
+ * TODO the reference value should be set from the build time, as well as being
+ * able to be changed via a flash setting.
+ */
+void gps_adjust_week_cycle(gps_time_t *t, u16 wn_ref) {
```
jacobmcnamee: Can we make this work regardless of whether or not t has already been adjusted?

ljbade:
@jacobmcnamee I added a check in case t was already adjusted, can you review?

src/time.c
jacobmcnamee: This still breaks if t->wn is less than wn_ref but not zero.

How about this:
```
if (t->wn < wn_ref)
  t->wn += 1024 * ((wn_ref + 1023 - t->wn) / 1024);
```

jacobmcnamee: Also as mentioned I think we should take in only the uncorrected week number and adopt the convention that gps_time_t only ever contains the corrected / absolute week number.

ljbade:
I changed the signature of `gps_adjust_week_cycle` to prevent accidental use with a `gpstime_t` that was already corrected. This function is only intended for use in `ephemeris.c`

jacobmcnamee:
Idk where my inline comments went, but how would you feel about changing to
```
if (wn_raw >= wn_ref)
  return wn_raw;

return wn_raw + 1024 * ((wn_ref + 1023 - wn_raw) / 1024);
```

ljbade:
@jacobmcnamee Done

PR #268, opened 2015-12-22
Fix up PRN codes

ljbade:
This PR inverts the GPS and SBAS PRN codes to match the definition of the ICDs.

It also reduces the SBAS PRN range to only the operational sats. Anything above PRN 138 is designated for testing only by the GPS PRN assignment plan.


ljbade:
I have tested this on two Piksis with RTK and everything works fine.

@swift-nav/firmware Ready for merge.

gsmcmullin:
Seems simple enough. Merge it.

Issue #217, opened 2015-09-09
Fixes to almanac.[ch]: entire GPS week number & gps_time structure


denniszollo:
Refactor almanac.c:
 * use gps_time_t to store wn and msec in week
 * require full week number for all function arguments and related documentation

Requires libsbp issue # and piksi_firmware #

PR #269, opened 2015-12-23
Improve almanac and ephemeris functions

ljbade:
I have re factored the almanac functions to reuse the ephemeris orbit calculations.

I have also implemented missing methods so now both ephemeris and almanac structs have an equivalent feature set.

I have also enhanced the documentation.

Remaining tasks:
- [x] add tests for new functions + old
- [x] fix up Python
- [x] in firmware change `manage.c` to use the new ephemeris az/el and Doppler functions


ljbade:
@denniszollo I think I will also incorporate the notes from your issue https://github.com/swift-nav/libswiftnav/issues/217

ljbade:
@denniszollo actually I think I have already addressed them. Do you mind reviewing what I have so far to see if they address your problems?

denniszollo:
@ljbade looking over my github notifications, I see I never looked at this. Do you still require some review?

ljbade:
@denniszollo hmm yeah but first i need to rebase and fix python

ljbade:
@denniszollo @swift-nav/estimation ready for review again

ljbade:
Travis failure looks like a Travis bug. The log shows everything passed.

ljbade:
Ready for review @swift-nav/estimation

ljbade:
Also @swift-nav/firmware can review

include/libswiftnav/ephemeris.h
```diff
@@ -18,47 +18,80 @@
...
-  gnss_signal_t sid;
-  gps_time_t toe;
-  float ura;
-  u8 fit_interval;
-  u8 valid;
-  u8 healthy;
+  gnss_signal_t sid; /**< Signal ID. */
+  gps_time_t toe;    /**< Reference time of ephemeris. */
+  float ura;         /**< User range accuracy [m] */
+  u8 fit_interval;   /**< Curve fit interval. */
```
rge-exafore: What are units?

ljbade: Oh good catch! Ill add them. URA is in metres, curve fit interval in hours

ljbade: My understanding is that GLONASS is 30mins? So shhould we make this a float?

ljbade:
Hold off on merging until me and @rge-exafore work out what units to use for `fit_interval`

include/libswiftnav/ephemeris.h
```diff
@@ -18,47 +18,80 @@
 #include <libswiftnav/time.h>
 #include <libswiftnav/common.h>
 
+/** \addtogroup ephemeris
+ * \{ */
+
+/** Structure containing the SBAS ephemeris for one satellite. */
```
rge-exafore: Should be 'GPS', not 'SBAS'

ljbade:
Still waiting for a review @swift-nav/firmware

src/ephemeris.c
```diff
@@ -25,56 +25,58 @@
...
-  pos[1] = ex->pos[1] + ex->rate[1] * dt +
+  pos[1] = ex->pos[1] + ex->vel[1] * dt +
            0.5 * ex->acc[1] * pow(dt, 2);
-  pos[2] = ex->pos[2] + ex->rate[2] * dt +
+  pos[2] = ex->pos[2] + ex->vel[2] * dt +
            0.5 * ex->acc[2] * pow(dt, 2);
 
-  memcpy(clock_err, &(ex->a_gf0), sizeof(double));
-  memcpy(clock_rate_err, &(ex->a_gf1), sizeof(double));
+  memcpy(clock_err, &(ex->a_gf0), sizeof(*clock_err));
```
jacobmcnamee: Is there a reason to use memcpy here instead of simple assignment?

src/almanac.c
```diff
@@ -26,208 +28,160 @@
...
-  /* Compute the satellite's velocity in Earth-Centered Earth-Fixed
-   * coordinates. */
-  if (vel) {
-    temp = y_dot * cos(alm->inc);
-    vel[0] = -om_dot * pos[1] + x_dot * cos(om) - temp * sin(om);
-    vel[1] =  om_dot * pos[0] + x_dot * sin(om) + temp * cos(om);
-    vel[2] = y_dot * sin(alm->inc);
-  }
-
+  ephemeris_t e;
```
jacobmcnamee: I'm not wild about copying all these parameters to a fake ephemeris struct, but I don't really see a better way. I suppose it's fine as long as it is thoroughly tested.

src/ephemeris.c
ljbade: No... just was how it was originally written. Will change to assignment.

jacobmcnamee:
Added a couple minor inline comments. Aside from that, it would be nice if we eventually checked the return codes from these functions. Otherwise lgtm.

src/almanac.c
ljbade: It's not really a fake ephemeris struct... the GPS almanac is supposed to be a subset of the full ephemeris, with just the missing parameters set to zero.

I wanted to ensure the same code for the satellite position is used for both ephemeris and almanac as they are supposed to be the same algorithm. It also prevents mistakes in the future where one function is changed but not the other.

jacobmcnamee: If that's the case, then it would suggest using a common structure for the shared parameters. I'm not sure we really want to do that, though.

Anyway reusing the code is great.

jacobmcnamee:
lgtm assuming hitl looks good

akleeman:
Looks like this change is breaking a `gnss_analysis` test, is [this](https://github.com/swift-nav/gnss-analysis/blob/master/tests/test_ephemeris.py#L37) still the proper way to call `calc_sat_state`?  Right now it returns all zeros for pos, vel, etc ...

PR #270, opened 2015-12-28
Fix almanac python binding init

benjamin0:
This changes the almanac initializer to be similar to the ephemeris python binding. The second commit has a failing test with the old initializer and the third commit has a passing test with the new initializer. 

@mookerji @fnoble

### DIFF
--- a/python/swiftnav/almanac.pyx
+++ b/python/swiftnav/almanac.pyx
@@ -26,8 +26,13 @@ cdef class Almanac:
 
   def __init__(self, **kwargs):
     memset(&self._thisptr, 0, sizeof(almanac_t))
-    if kwargs:
-      self._thisptr = kwargs
+    self._thisptr.sid = kwargs.pop('sid')
+    self._thisptr.healthy = kwargs.pop('healthy')
+    self._thisptr.valid = kwargs.pop('valid')
+    if 'gps' in kwargs:
+      self._thisptr.gps = kwargs.pop('gps')
+    elif 'sbas' in kwargs:
+      self._thisptr.sbas = kwargs.pop('sbas')
 
   def __getattr__(self, k):
     return self._thisptr.get(k)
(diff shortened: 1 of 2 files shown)


ljbade:
How will https://github.com/swift-nav/libswiftnav/pull/269 affect this?

python/swiftnav/almanac.pyx
```diff
@@ -111,6 +116,6 @@ cdef class Almanac:
       The Doppler shift in Hz.
 
     """
-    assert len(ref) != 3, "ECEF coordinates must have dimension 3."
+    assert len(ref) == 3, "ECEF coordinates must have dimension 3."
```
mookerji: Derp. I'm an idiot.

python/tests/test_almanac.py
```diff
@@ -9,6 +9,7 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
+import numpy as np
 import swiftnav.almanac
 
 def test_imports():
```
mookerji: I'd remove this `test_imports` thing up here.

python/tests/test_almanac.py
```diff
@@ -16,3 +17,32 @@ def test_imports():
 
   """
   assert True
+
+def test_init():
```
mookerji: Great.

Would not mind seeing a quick sanity test for the calc functions, which I think you can get from an ephemeris definition?

mookerji:
Sorry this took so long, I had just a nit. If you can add another test case or two, that would be sweet. Otherwise, merge whenever.

python/tests/test_almanac.py
benjamin0: I figured functional tests of the almanac library would be found in the C tests/ folder whereas these tests would test the python bindings, but now that I look there are no C almanac tests. I'll add a simple test here, but I know @ljbade has #269 which merges the almanac and ephemeris code so that should improve almanac/ephemeris test coverage.

benjamin0:
@mookerji Added some sanity tests, if it looks good, I'll merge

PR #272, opened 2016-01-07
Improve UTC handling (WIP)

ljbade:
Main improvements:
- replace using system Unix time functions (which introduces many problems) with our own
- convert GPS network time into UTC

So far the conversion of seconds into a calendar data and clock display works perfectly for the next 32,000 years with an extensive unit test.

However the leap second event handling is broken. It has something to do with the GPS ICD's algorithim returns seconds modulo a day, which make it hard to roll over the date when the leap second happens at midnight.

I have not found any public or open source implementation of GPS leap seconds to study.


ljbade:
Making a note to check gpstk code

ljbade:
Making a note `gps2doy` should also be changed to use these functions.

PR #273, opened 2016-01-07
Improve troposphere model (WIP)

ljbade:
Working on implementing the UNB3m model from http://gauss.gge.unb.ca/papers.pdf/ionntm2006.leandro.pdf

It also will use the Niell mapping functions from http://www.haystack.mit.edu/geo/pubs/NMF_JGR.pdf or the UNBabc if Niell is too expensive http://gauss2.gge.unb.ca/papers.pdf/iongpsgnss2003.guo.pdf

This is an improvement to fix issues in original UNB3 which is what is recommended by the SBAS specifications.

I believe this to be the best model that does not require large memory/CPU overhead and is suitable for offline implementation on v2.

In the future we may want to move to GPT2 or UMB-VMF1 models.


ljbade:
@henryhallam Do you think I have selected a good troposphere model to implement?

ljbade:
Closed in favour of https://github.com/swift-nav/libswiftnav/pull/343

PR #274, opened 2016-01-08
Minor vector length bug in dgnss_management

akleeman:
The call to `make_measurements` was passing the number of sdiffs, not the number of double diffs and the vector passed along was too small (causing seg faults).

@mookerji

### DIFF
--- a/python/swiftnav/dgnss_management.pyx
+++ b/python/swiftnav/dgnss_management.pyx
@@ -34,11 +34,11 @@ def dgnss_set_settings_(phase_var_test, code_var_test,
                      amb_drift_var, amb_init_var,new_int_var)
 
 def make_measurements_(sdiffs):
-  num_sdiffs = len(sdiffs)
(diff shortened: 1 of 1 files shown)


ljbade:
add to whitelist

ljbade:
@akleeman not sure if you were told by @fnoble yet, but we now have GitHub teams so you can @swift-nav/firmware if you think more than one person is capable of reviewing the PR. This should help get PRs review faster.

akleeman:
cool, I'll definitely use those in the future.

@ljbade: What did you mean by add to whitelist?

ljbade:
@akleeman it is a special message that allow Jenkins to pick up your pr and build it. You are now on the white list so next PRs you make Jenkins will automatically accept

mookerji:
@ljbade I think travis is also building PRs here as well, and it might be a bit more up-to-date.

PR #277, opened 2016-01-08
Move almanac util file from sbp_log_analysis to libswiftnav.

benjamin0:
Buro suggested (https://github.com/swift-nav/sbp_log_analysis/pull/288) moving this file here so sbp_log_analysis and piksi_tools can both use it.

@mookerji @fnoble


benjamin0:
@fnoble recommended that we keep the almanac util file in sbp_log_analysis so I'm going to close this.

PR #282, opened 2016-01-15
Python time bug

denniszollo:
I think this fixes some bugs in the python bindings around gpstime.  I also added a #! to the setup.py

I'm not 100% sure of the intended architecture in the python bindings, but this allowed me to use them.
/cc @mookerji @akleeman 

Next steps are to allow cython to work with a local libswiftnav library.

### DIFF
--- a/python/swiftnav/time.pyx
+++ b/python/swiftnav/time.pyx
@@ -16,8 +16,8 @@ cdef class GpsTime:
 
   def __init__(self, **kwargs):
     memset(&self._thisptr, 0, sizeof(gps_time_t))
-    self._thisptr.tow = kwargs.pop('tow')
-    self._thisptr.wn = kwargs.pop('wn')
+    if kwargs:
+      self._thisptr = kwargs
 
   def __getattr__(self, k):
     return self._thisptr.get(k)
--- a/python/swiftnav/time.pyx
+++ b/python/swiftnav/time.pyx
@@ -16,9 +16,12 @@ cdef class GpsTime:
 
   def __init__(self, **kwargs):
     memset(&self._thisptr, 0, sizeof(gps_time_t))
-    if kwargs:
-      self._thisptr = kwargs
+    self._thisptr.tow = kwargs.pop('tow')
+    self._thisptr.wn = kwargs.pop('wn')
 
+  def __getattr__(self, k):
+    return self._thisptr.get(k)
+  
   def __repr__(self):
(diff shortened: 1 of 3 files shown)


benjamin0:
This looks similar to my PR: https://github.com/swift-nav/libswiftnav/pull/270/ for fixing the almanac. Maybe we should go through and change all the initializers?

ljbade:
@benjamin0 that would be sensible

python/swiftnav/time.pyx
```diff
@@ -16,9 +16,12 @@ cdef class GpsTime:
 
   def __init__(self, **kwargs):
     memset(&self._thisptr, 0, sizeof(gps_time_t))
-    if kwargs:
-      self._thisptr = kwargs
+    self._thisptr.tow = kwargs.pop('tow')
```
mookerji: Errr, what? Are you sure this doesn't work? There are definitely tests that show that this initializer should work. You have to pass a "**"'d dict to the constructor, like `GpsTime(**{'wn': ...})`.

mookerji:
@benjamin0 @ljbade the initializers are like this by design: the goal is to reduce initializer boilerplate. If you pass in a dictionary in Cython, it'll initialize a C struct using its a contents with a copy. In fact, I somehow didn't notice this when I was looking at #270: that almanac initializer *should* work the same way.

python/swiftnav/time.pyx
```diff
@@ -97,4 +100,4 @@ def datetime2gpst(timestamp):
   wn = dt.days / 7
   tow = (dt - datetime.timedelta(weeks=wn)).total_seconds()
   #wn % 1024 would be standard, but we want a bijective tranformation.
-  return GpsTime(wn, tow)
+  return GpsTime(wn=wn, tow=tow)
```
mookerji: Yeah, this is definitely a bug.

benjamin0:
@mookerji I was having an issue where some of the parameter values weren't the same after being passed through the initializer (specifically `gps.ecc`). Try running the tests between https://github.com/swift-nav/libswiftnav/commit/3041ad267d7595052cf27d068bf9ef14ccf704fb and https://github.com/swift-nav/libswiftnav/commit/00cd67ab1a13990a31da0f20efea2ae1edf0bd8c and let me know if you run into the same issue.

python/swiftnav/time.pyx
denniszollo: Now that I know how the initializers are supposed to work, I'll fixup this PR.

denniszollo:
@mookerji 

Now I think this is ready for merging.

PR #284, opened 2016-01-16
Print navigation measurements, bug fix, and test.

mookerji:
Notably, adds a test case for the make_propagated_sdiffs function with
some synthetic data taken from Kleeman's SITL runner. The case is
somewhat huge, so I think we'd ultimately want to replace something
like this something that could be generated more cleanly within the
library.

/cc @akleeman @benjamin0 @fnoble

### DIFF
--- a/python/swiftnav/ambiguity_test.pyx
+++ b/python/swiftnav/ambiguity_test.pyx
@@ -137,9 +137,7 @@ cdef class AmbiguityTest:
                             is_bad_measurement):
     num_sdiffs = len(sdiffs)
     cdef sdiff_t sdiffs_[32]
-    print "here"
     mk_sdiff_array(sdiffs, 32, &sdiffs_[0])
-    print sdiffs_
     return ambiguity_update_sats(&self._thisptr,
                                  num_sdiffs,
                                  &sdiffs_[0],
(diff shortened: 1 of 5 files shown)


mookerji:
Errrrr, not quite right. Will adjust a bit later today.

fnoble:
Could we put the bug fix in a separate PR so we can get it in right away? Also I'd vote for dropping this test for now, it compares a function that we know we want to change with one that is in itself untested. Not sure it really verifies anything.

ljbade:
Yeah split out the bug fix for now.

On Fri, Jan 15, 2016 at 7:11 PM, Fergus Noble <notifications@github.com>
wrote:

> Could we put the bug fix in a separate PR so we can get it in right away?
> Also I'd vote for dropping this test for now, it compares a function that
> we know we want to change with one that is in itself untested. Not sure it
> really verifies anything.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/swift-nav/libswiftnav/pull/284#issuecomment-172150169>
> .
>



-- 
Thanks,
Leith Bade
leith@swiftnav.com

akleeman:
You could make a smaller test that simply made sure that the returned sdiffs are non-zero (which was the original bug).

mookerji:
Ok, I reduced the test case just to sanity check the outputs, per @akleeman's recommendation.

PR #290, opened 2016-01-22
GPS CNAV message decoding support

valeri-atamaniouk:
Added capability to decode GPS CNAV message and extract the following information: CNAV message header, symbol delay.


jrostrom:
The comment in _cnav_msg_invert() mentions "The data inversion flag is also updated". But there doesn't seem up be an update in the function. Is the intention that it's done elsewhere?

PR #291, opened 2016-01-22
GPS CNAV message decoding support

valeri-atamaniouk:
Added capability to decode GPS CNAV message and extract the following information: CNAV message header, symbol delay.

### DIFF
--- a/include/libswiftnav/cnav_msg.h
+++ b/include/libswiftnav/cnav_msg.h
@@ -1,0 +1,99 @@
+/*
+ * Copyright (C) 2016 Swift Navigation Inc.
+ * Contact: Valeri Atamaniouk <valeri@swift-nav.com>
+ *
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef LIBSWIFTNAV_CNAV_MSG_H
+#define LIBSWIFTNAV_CNAV_MSG_H
+
+#include <libswiftnav/common.h>
+#include <libfec/fec.h>
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <limits.h>
+
+/** \addtogroup GPS_L2
+ * \{ */
+/** \addtogroup gps_cnav_decoder
+ * \{ */
+
+/** Size of the Viterbi decoder history. */
+#define GPS_L2_V27_HISTORY_LENGTH_BITS 64
+/** Bits to accumulate before decoding starts. */
+#define GPS_L2C_V27_INIT_BITS    (32)
+/** Bits to decode at a time. */
+#define GPS_L2C_V27_DECODE_BITS  (32)
+/** Bits in decoder tail. We ignore them. */
+#define GPS_L2C_V27_DELAY_BITS   (32)
+/**
+ * GPS CNAV message container.
+ *
+ * @sa cnav_msg_decoder_add_symbol
+ */
+typedef struct
+{
+  u8   prn;    /**< SV PRN. 0..31 */
+  u8   msg_id; /**< Message id. 0..31 */
+  u32  tow;    /**< GPS ToW in 6-second units. Multiply to 6 to get seconds. */
+  bool alert;  /**< CNAV message alert flag */
+} cnav_msg_t;
+
+/**
+ * GPS CNAV decoder component.
+ * This component controls symbol decoding string.
+ *
+ * @sa cnav_msg_decoder_t
+ */
+typedef struct {
+  v27_t          dec;           /**< Viterbi block decoder object */
+  v27_decision_t decisions[GPS_L2_V27_HISTORY_LENGTH_BITS];
+                                /**< Decision graph */
+  unsigned char  symbols[(GPS_L2C_V27_INIT_BITS + GPS_L2C_V27_DECODE_BITS) * 2];
+                                /**< Symbol buffer */
+  size_t         n_symbols;     /**< Count of symbols in the symbol buffer */
+  unsigned char  decoded[GPS_L2C_V27_DECODE_BITS + GPS_L2C_V27_DELAY_BITS];
+                                /**< Decode buffer */
+  size_t         n_decoded;     /**< Number of bits in the decode buffer */
+  bool           preamble_seen; /**< When true, the decode buffer is aligned on
+                                 *   preamble. */
+  bool           invert;        /**< When true, indicates the bits are inverted */
+  bool           message_lock;  /**< When true, indicates the message boundary
+                                 *   is found. */
+  bool           crc_ok;        /**< Flag that the last message had good CRC */
+  size_t         n_crc_fail;    /**< Counter for CRC failures */
+  bool           init;          /**< Initial state flag. When true, initial bits
+                                 *   do not produce output. */
+} cnav_v27_part_t;
+
+/**
+ * GPS CNAV message lock and decoder object.
+ *
+ * Decoder uses two Viterbi decoder objects to ensure the lock is acquired when
+ * the input symbol phase is not known.
+ */
+typedef struct
+{
+  cnav_v27_part_t part1; /**< Decoder for odd symbol pairs */
+  cnav_v27_part_t part2; /**< Decoder for even symbol pairs */
+} cnav_msg_decoder_t;
+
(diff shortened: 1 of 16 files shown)


rge-exafore:
@swift-nav/firmware, could you please review and merge?

src/cnav_msg.c
```diff
@@ -0,0 +1,449 @@
...
+ * CNAV decoder contains two Viterbi decoders that are used to estimate bit and
+ * message boundary.
+ *
+ * \param[out] dec Decoder structure.
+ *
+ * \return None
+ */
+void cnav_msg_decoder_init(cnav_msg_decoder_t *dec)
+{
+  static const signed char polynomial[2] = {
```
jacobmcnamee: polynomial does not need to be statically allocated

src/cnav_msg.c
```diff
@@ -0,0 +1,449 @@
...
+   * - M - Number of bits in the tail to ignore.
+   */
+  unsigned char tmp_bits[ (GPS_L2C_V27_DECODE_BITS + GPS_L2C_V27_DELAY_BITS +
+                           CHAR_BIT - 1) / CHAR_BIT];
+
+  v27_chainback_likely(&part->dec, tmp_bits,
+                       GPS_L2C_V27_DECODE_BITS + GPS_L2C_V27_DELAY_BITS);
+
+  /* Read decoded bits and add them to the decoded buffer */
+  u32 to_add = getbitu(tmp_bits, 0, GPS_L2C_V27_DECODE_BITS);
```
jacobmcnamee: This breaks if GPS_L2C_V27_DECODE_BITS > 32. Can we add a comment or an assert or something?

src/cnav_msg.c
valeri-atamaniouk: I disagree. The coefficients never change and they have to be passed as an array. There is no reason to allocate they array in heap and programmatically fill it.

jacobmcnamee: I mean it is not required - the Viterbi decoder will work fine either way.

If we're concerned with speed or memory, it might be good to preinitialize and share a single v27_poly_t between all the CNAV decoders.

jacobmcnamee:
A couple // comments in the header file and test file, but otherwise lgtm.

include/libswiftnav/cnav_msg.h
```diff
@@ -0,0 +1,95 @@
...
+  u32  tow;    //!< GPS ToW in 6-second units. Multiply to 6 to get seconds.
+  bool alert;  //!< CNAV message alert flag
+} cnav_msg_t;
+
+/**
+ * GPS CNAV decoder component.
+ * This component controls symbol decoding string.
+ *
+ * @sa cnav_msg_decoder_t
+ */
```
fnoble: Comments and documentation are great! Just a small note that we use a slightly different style for our Doxygen comments. Not a big deal but good to be consistent.

https://swiftnav.hackpad.com/Engineering-Playbook-fKcJhOl85sD#:h=Doxygen

src/cnav_msg.c
```diff
@@ -0,0 +1,515 @@
...
+/** \defgroup GPS_L2 GPS L2
+ * GPS L2 operations
+ * \{ */
+/** \defgroup gps_cnav_decoder Decoder
+ * GPS L2C CNAV message decoding.
+ * \{ */
+/*
+ * Block Viterbi decoding parameters.
+ */
+/** Viterbi decoder inverted polynome A */
```
fnoble: polynome -> polynomial

src/cnav_msg.c
```diff
@@ -0,0 +1,515 @@
...
+/** GPS L2C CNAV message length in bits */
+#define GPS_CNAV_MSG_LENGTH         (300)
+/** GPS LC2 CNAV CRC length in bits */
+#define GPS_CNAV_MSG_CRC_LENGTH     (24)
+/** GPS L2C CNAV message payload length in bits */
+#define GPS_CNAV_MSG_DATA_LENGTH    (GPS_CNAV_MSG_LENGTH-GPS_CNAV_MSG_CRC_LENGTH)
+/** GPS L2C CNAV message lock detector threshold */
+#define GPS_CNAV_LOCK_MAX_CRC_FAILS (10)
+
+// forward declarations for test access
```
fnoble: `//` -> `/* */`

tests/check_cnav.c
```diff
@@ -0,0 +1,389 @@
...
+#include <stdlib.h>
+#include <math.h>
+#include <check.h>
+
+#include "check_utils.h"
+
+#define G1 0x79 // 0171
+#define G2 0x5B // 0133
+
+// internal CNAV declarations for test access
```
fnoble: I worry this could cause unexpected behavior if you forget to keep the prototypes in sync. Another option is to declare the functions static in `cnav_msg.c` and then in this test file do a `#include <cnav_msg.c>` (note .c not .h !)

Anyway, I don't mind too much which solution you choose. This stack overflow post has more ideas:
http://stackoverflow.com/questions/593414/how-to-test-a-static-function

src/edc.c
```diff
@@ -77,6 +77,46 @@ u32 crc24q(const u8 *buf, u32 len, u32 crc)
   return crc;
 }
 
+/**
+ * Computes CRC-24Q for left-aligned bit message.
+ * This function is used for left-aligned bit messages, for example SBAS and
+ * GPS CNAV.
+ * GPS message is 300 bits total, but 276 bits without CRC. It takes 34.5
+ * 8-bit bytes, and when computing CRC the message has to be padded with zero
+ * bits.
```
fnoble: @jacobmcnamee : how did you deal with this for SBAS? Did you use the existing CRC function?

src/cnav_msg.c
```diff
@@ -0,0 +1,515 @@
...
+ *
+ * \param[in] buf   Pointer to buffer head.
+ * \param[in] bits  Number of bytes in the buffer.
+ * \param[in] shift Number of bits for left shift operation.
+ *
+ * \return None
+ *
+ * \private
+ */
+void _cnav_bitlshift(void *buf, size_t size, u32 shift)
```
fnoble: This function seems generally useful outside of CNAV, maybe move to `bits.c`?

src/cnav_msg.c
```diff
@@ -0,0 +1,515 @@
...
+ * \param[in]     src       Source buffer.
+ * \param[in]     src_index Source bit index.
+ * \param[in]     count     Number of bits to copy.
+ *
+ * \return None
+ *
+ * \private
+ */
+void _cnav_bitcopy(void *dst, u32 dst_index,
+                   const void *src, u32 src_index, u32 count)
```
fnoble: Also seems generally useful, move to `bits.c`?

src/cnav_msg.c
```diff
@@ -0,0 +1,515 @@
...
+/**
+ * Invert message bits in the buffer.
+ *
+ * The method inverts bits of the decoded data.
+ *
+ * \param[in,out] part Decoder component with a message buffer.
+ *
+ * \return None
+ */
+void _cnav_msg_invert(cnav_v27_part_t *part)
```
fnoble: Also a candidate to be turned into a generic invert buffer function in `bits.c`?

src/edc.c
jacobmcnamee: We didn't really get that far. Vlad's SBAS support changes for piksi_firmware or libswiftnav might have something if you are really curious.

src/cnav_msg.c
```diff
@@ -0,0 +1,515 @@
...
+  dec->part2.init = true;
+  _cnav_add_symbol(&dec->part2, 0x80);
+}
+
+/**
+ * Adds a received symbol to decoder.
+ *
+ * \param[in,out] dec    Decoder object.
+ * \param[in]     symbol Symbol value probability, where 0x00 - 100% of 0,
+ *                       0xFF - 100% of 1.
```
fnoble: Might want to double check, I thought the `symbol` was passed as a signed number, i.e. 2's complement

tests/check_cnav.c
```diff
@@ -0,0 +1,389 @@
...
+
+#include <time.h>
+#include <stdlib.h>
+#include <math.h>
+#include <check.h>
+
+#include "check_utils.h"
+
+#define G1 0x79 // 0171
+#define G2 0x5B // 0133
```
fnoble: `//` -> `/* */` in a few places

fnoble:
@valeri-atamaniouk : looking really good! I made a few comments but mostly minor.

I don't know if you have seen this but we have code coverage testing available here:
https://coveralls.io/builds/4852544/source?filename=src%2Fcnav_msg.c

Looks like there are a couple edge conditions currently missed that it would be good to hit.

Just address those comments and test coverage and we'll be good to merge.

include/libswiftnav/cnav_msg.h
valeri-atamaniouk: Docs updated. No more C++ - style comments / documentation blocks

valeri-atamaniouk:
Please do not merge. Some newer tests don't pass.

valeri-atamaniouk:
@fnoble All comments addressed: code refactored, new unit-tests, coverage ok.

PR #293, opened 2016-01-27
Split ephemeris_valid into ephemeris_valid_at_time

ljbade:
This is so code that just wants to check if a valid ephemeris has been decoded (and not caring about it's age), can use a function rather than directly checking `ephemeris_t->valid` for consistency


ljbade:
@swift-nav/firmware ready for review

ljbade:
Closing in favour of https://github.com/swift-nav/libswiftnav/pull/304

PR #295, opened 2016-01-28
Add L2CM support to signal module

pmiettinen:
Original indexing idea is preserved with added functionality
to support the possibility of a satellite containing multiple
signals.


pmiettinen:
Design https://swiftnav.hackpad.com/Signal.ch-L2C-update-jAKrz02zkU9

pmiettinen:
@fnoble @jacobmcnamee @gsmcmullin please review

src/signal.c
```diff
@@ -15,67 +16,61 @@
...
+  return sid;
 }
 
 int sid_to_string(char *s, int n, gnss_signal_t sid)
 {
   const char *constellation_str = (sid.constellation < CONSTELLATION_COUNT) ?
       constellation_strs[sid.constellation] : unknown_str;
-  const char *band_str = (sid.band < BAND_COUNT) ?
-      band_strs[sid.band] : unknown_str;
+  const char *band_str = (sid.code < CODE_COUNT) ?
```
gsmcmullin: I'd prefer to remove the CODE_COUNT which now has to be manually defined, and change this to a switch statement on the enum type.

pmiettinen: I agree on the switch statement addition, but what's the option for CODE_COUNT manual update?

pmiettinen:
@fnoble @jacobmcnamee @gsmcmullin please review

pmiettinen:
@cbeighley please review

include/libswiftnav/signal.h
```diff
@@ -71,12 +94,17 @@ int sid_to_string(char *s, int n, gnss_signal_t sid);
...
+gnss_signal_t sid_from_index(u16 i);
+
+/** Converts sid to a global index in [0, NUM_CODES) if INDEX_ALL_SIGNALS and
+ *  to constellation index in [0, NUM_SATS_CONS) if INDEX_SATS_IN_CONS
  */
-gnss_signal_t sid_from_index(u32 i);
+u16 sid_to_index(gnss_signal_t sid, enum indexing_type it);
 
-/** Converts sid to a global index in [0, NUM_SATS)
+/** Returns the sid in which the sid belongs to.
```
ljbade: this doesnt make sense

include/libswiftnav/signal.h
```diff
@@ -37,16 +58,14 @@ enum band {
 
 typedef struct __attribute__((packed)) {
   u16 sat;
-  u8 band;
-  u8 constellation;
+  enum code code;
+  u8 reserved;
```
ljbade: Do we need to keep this field?

src/bit_sync.c
```diff
@@ -41,8 +42,8 @@ void bit_sync_init(bit_sync_t *b, gnss_signal_t sid)
   memset(b, 0, sizeof(bit_sync_t));
   b->bit_phase_ref = BITSYNC_UNSYNCED;
 
-  assert(sid.band == BAND_L1);
-  switch (sid.constellation) {
+  assert(sid.code == CODE_GPS_L1CA || sid.code == CODE_SBAS_L1CA);
```
ljbade: @jacobmcnamee should this also have sbas?

src/signal.c
```diff
@@ -15,113 +16,171 @@
...
+  return sid;
 }
 
 int sid_to_string(char *s, int n, gnss_signal_t sid)
 {
-  const char *constellation_str = (sid.constellation < CONSTELLATION_COUNT) ?
-      constellation_strs[sid.constellation] : unknown_str;
-  const char *band_str = (sid.band < BAND_COUNT) ?
-      band_strs[sid.band] : unknown_str;
+  const char *code_str = unknown_str;
```
ljbade: Why was this changed to use switch instead of index of strings?

src/signal.c
```diff
@@ -15,113 +16,171 @@
...
+
+  sid.sat = (idx - first_sat_index[cons]) /
+      codes_on_cons[cons] + first_sat_id[cons];
+
   return sid;
 }
 
-u32 sid_to_index(gnss_signal_t sid)
+/** Return local index based on code.
+ *  Example if (side.code == CODE_GPS_L2CM) return 1;
```
ljbade: should be `sid` instead of `side`

src/signal.c
pmiettinen: At that point we had different approach for the code enum. I'll change this back to table of strings.

include/libswiftnav/signal.h
gsmcmullin: I think this reserved field can be removed, as can the packed attribute.

jacobmcnamee: Agree. We have different structs and converter functions for internal and external (SBP message) representations. Internal representation shouldn't matter.

include/libswiftnav/signal.h
```diff
@@ -19,15 +20,35 @@
 #define NUM_SATS_SBAS 19
 #define NUM_SATS (NUM_SATS_GPS + NUM_SATS_SBAS)
 
+#define CODES_ON_GPS_SAT 2
+#define CODES_ON_SBAS_SAT 1
+
+#define NUM_CODES_GPS (NUM_SATS_GPS * CODES_ON_GPS_SAT)
```
jacobmcnamee: I think NUM_CODES_GPS should be the number of different types of codes on GPS. So L1CA + L2CM + L2CL = 3 ?

NUM_SIGNALS_GPS should be NUM_SATS_GPS * NUM_CODES_GPS

include/libswiftnav/signal.h
```diff
@@ -61,6 +80,10 @@ static inline bool sid_is_equal(const gnss_signal_t a, const gnss_signal_t b)
   return sid_compare(a, b) == 0;
 }
 
+/** Constructs and returns a gnss_signal_t structure using given parameters.
+ */
+gnss_signal_t construct_sid(enum code code, u16 sat);
```
jacobmcnamee: Do we need this function for something in particular? It requires the caller to know in advance the valid sat / code combinations, which I was hoping to abstract away in signal.c

include/libswiftnav/signal.h
```diff
@@ -71,12 +94,17 @@ int sid_to_string(char *s, int n, gnss_signal_t sid);
  */
 bool sid_valid(gnss_signal_t sid);
 
-/** Converts the global index i in [0, NUM_SATS) to the corresponding sid
+/** Converts the global index i in [0, NUM_CODES) to the corresponding sid
```
jacobmcnamee: [0, NUM_SIGNALS)

include/libswiftnav/signal.h
pmiettinen: Sounds reasonable, although at the moment L2CL is not really supported, it just ended up into the code enum for some reason. Maybe it should be still omitted at this point.

include/libswiftnav/signal.h
pmiettinen: @rge-exafore did you have some use case for this?

include/libswiftnav/signal.h
pmiettinen: @cbeighley do you want to comment on this since you voted for to keep reserved field?

include/libswiftnav/signal.h
```diff
@@ -19,15 +20,35 @@
...
+enum code {
+  CODE_INVALID = -1,
+  CODE_GPS_L1CA,
+  CODE_GPS_L2CL,
+  CODE_GPS_L2CM,
+  CODE_SBAS_L1CA,
+  CODE_COUNT,
+};
+
+enum indexing_type {
```
jacobmcnamee: I think we want
INDEX_GLOBAL [0, NUM_SIGNALS)
INDEX_CONSTELLATION [0, NUM_SIGNALS_<constellation>)
INDEX_CODE [0, NUM_SIGNALS_<code>)

src/bit_sync.c
jacobmcnamee: For now, yes, those are the signal types with defined bit lengths (see top of file). Eventually we should store the bit lengths for all codes in some central location and remove the asserts.

src/bit_sync.c
```diff
@@ -41,8 +42,8 @@ void bit_sync_init(bit_sync_t *b, gnss_signal_t sid)
...
+  switch (sid_to_constellation(sid)) {
```
jacobmcnamee: This should just switch on code

src/signal.c
```diff
@@ -15,113 +16,162 @@
...
+  return sid;
 }
 
 int sid_to_string(char *s, int n, gnss_signal_t sid)
 {
-  const char *constellation_str = (sid.constellation < CONSTELLATION_COUNT) ?
-      constellation_strs[sid.constellation] : unknown_str;
-  const char *band_str = (sid.band < BAND_COUNT) ?
-      band_strs[sid.band] : unknown_str;
+  const char *code_str = (sid.code < CODE_COUNT &&
```
jacobmcnamee: This breaks if sid.code is CODE_INVALID = -1

src/signal.c
```diff
@@ -15,113 +16,162 @@
 
 #include <libswiftnav/signal.h>
 
-typedef struct {
-  u8 constellation;
-  u8 band;
-  u16 sat_count;
-  u16 (*sat_get)(u32 local_index);
-  u32 (*local_index_get)(u16 sat);
```
jacobmcnamee: FYI the idea with the local_index_get() stuff was that we could have non-contiguous blocks of sat values for a given code. But I think we should keep your changes until we need something more sophisticated (if that ever happens).

src/signal.c
pmiettinen: True

include/libswiftnav/signal.h
rge-exafore: Yes, when we start tracking for L2C from acquisition on L1CA we need to construct new sid and we know sat and code. So, for cases like this.

include/libswiftnav/signal.h
cbeighley: Agree with Gareth and Jacob, let's remove it.

include/libswiftnav/signal.h
cbeighley: Seems reasonable. @jacobmcnamee did you have a different idea about how to do this?

include/libswiftnav/signal.h
jacobmcnamee: What matters is the SBP representation. We should keep the extra byte there, but here it doesn't matter.

include/libswiftnav/signal.h
pmiettinen: What's the idea behind this, wouldn't this require also constellation and code as parameter?

include/libswiftnav/signal.h
jacobmcnamee: Nothing significantly better for that use case. I think it's fine as long as we only use it where necessary. (For example, NOT to initialize a huge list of SIDs)

jacobmcnamee:
Looks good to me assuming the tests are passing and it's working as intended.

jacobmcnamee:
Superseded by #303

PR #297, opened 2016-02-09
Refactor dgnss return codes, and make nav phase error a warning

ljbade:
Change error message "Nav phase flip - half cycle slip detected" to a warning.

Resolve ambigous -3 return in `lesq_solve_raim`.


ljbade:
@swift-nav/firmware Ready for review

src/baseline.c
```diff
@@ -501,7 +501,7 @@ s8 lesq_solve_raim(u8 num_dds_u8, const double *dd_obs,
     if (n_used) {
       *n_used = 0;
     }
-    return -3;
+    return -5;
```
gsmcmullin: Add this return value in the Doxygen comment.  This should probably be changed to an enum type rather than having these magic constants.

ljbade: Yeah I'm not a fan of the magic constants either.

It should also print out the meaning of the value too, so perhaps a function to return an string from the enum will be good.

ljbade:
Noting the first two changes (error to warning, and changing the -3 dgnss return value) are in https://github.com/swift-nav/libswiftnav/pull/310.

Will leave this PR to look at changing the return codes into enums

PR #299, opened 2016-02-11
Improve calc_sat_state error checking

ljbade:
Also fixes a compile error in `bitshl`

### DIFF
--- a/src/ephemeris.c
+++ b/src/ephemeris.c
@@ -41,15 +41,6 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
                              double pos[3], double vel[3],
                              double *clock_err, double *clock_rate_err)
 {
-  /*
-   *Ephemeris did not get time-stammped when it was received.
-   */
-  if (e->toe.wn == 0)
-    return -1;
-
-  if (!e->valid || !e->healthy)
-    return -1;
-
   const ephemeris_xyz_t *ex = &e->xyz;
   u8 ndays = t->tow / DAY_SECS;
   double tod = t->tow - DAY_SECS * ndays;
@@ -110,12 +101,6 @@ static s8 calc_sat_state_kepler(const ephemeris_t *e,
 
   /* Seconds from the time from ephemeris reference epoch (toe) */
   dt = gpsdifftime(t, &e->toe);
-
-  /* If dt is greater than fit_interval hours our ephemeris isn't valid. */
-  if (fabs(dt) > ((u32)e->fit_interval)*60*60) {
-    log_error("Using ephemeris outside validity period, dt = %+.0f", dt);
-    return -1;
-  }
 
   /* Calculate position per IS-GPS-200D p 97 Table 20-IV */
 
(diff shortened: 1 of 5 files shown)


src/ephemeris.c
```diff
@@ -253,9 +245,22 @@ u8 ephemeris_valid(const ephemeris_t *eph, const gps_time_t *t)
   /* Seconds from the time from ephemeris reference epoch (toe) */
   double dt = gpsdifftime(t, &eph->toe);
 
+  /*
+   *Ephemeris did not get time-stammped when it was received.
+   */
+  if (eph->toe.wn == 0) {
+    return 0;
```
jacobmcnamee: -1?

src/ephemeris.c
```diff
@@ -253,9 +245,22 @@ u8 ephemeris_valid(const ephemeris_t *eph, const gps_time_t *t)
...
+  }
+
   /* TODO: this doesn't exclude ephemerides older than a week so could be made
    * better. */
-  return (eph->valid && fabs(dt) < ((u32)eph->fit_interval)*60*60);
+  /* If dt is greater than fit_interval hours our ephemeris isn't valid. */
+  if (fabs(dt) > ((u32)eph->fit_interval) * 60 * 60) {
+    log_error("Using ephemeris outside validity period, dt = %+.0f", dt);
+    return 0;
```
jacobmcnamee: -1?

src/ephemeris.c
```diff
@@ -253,9 +245,22 @@ u8 ephemeris_valid(const ephemeris_t *eph, const gps_time_t *t)
...
+   *Ephemeris did not get time-stammped when it was received.
+   */
+  if (eph->toe.wn == 0) {
+    return 0;
+  }
+
   /* TODO: this doesn't exclude ephemerides older than a week so could be made
    * better. */
-  return (eph->valid && fabs(dt) < ((u32)eph->fit_interval)*60*60);
+  /* If dt is greater than fit_interval hours our ephemeris isn't valid. */
```
jacobmcnamee: I don't think we should warn in this function if the ephemeris is invalid. It's ok to pass outdated ephemeris to this function.

jacobmcnamee: Perhaps not the ideal place for it, but maybe just warn if fit_interval is zero?

jacobmcnamee: Also log_warn() either way

src/ephemeris.c
```diff
@@ -250,12 +248,28 @@ s8 calc_sat_state(const ephemeris_t *e, const gps_time_t *t,
  */
 u8 ephemeris_valid(const ephemeris_t *eph, const gps_time_t *t)
 {
+  assert(eph != NULL);
+  assert(t != NULL);
+  assert(eph->fit_interval > 0);
```
jacobmcnamee: I don't think this should assert on fit interval. Maybe just warn if everything else is ok. I believe all fields zero with valid = false needs to be tolerated (but obviously return false).

src/ephemeris.c
```diff
@@ -228,7 +213,20 @@ s8 calc_sat_state(const ephemeris_t *e, const gps_time_t *t,
   assert(clock_err != NULL);
   assert(clock_rate_err != NULL);
   assert(e != NULL);
-  assert(ephemeris_valid(e, t));
+
+  /* If dt is greater than fit_interval hours our ephemeris isn't valid. */
+  double dt = gpsdifftime(t, &e->toe);
```
jacobmcnamee: This code is duplicated in ephemeris_valid()

ljbade: Yes but it was the only way to get a warning in here but not in ephemeris_valid about using the old ephemeris. Maybe I will just update the other warning to indicate it could be either in invalid ephemeris, or a too old ephemeris.

jacobmcnamee:
Lgtm. Please merge.

PR #300, opened 2016-02-11
update requirements.txt to require Cython>=0.22

StefanD986:
As told in the readme Cython needs to be >=0.22, otherwise the build fails.


ljbade:
@mookerji Is this change right?

StefanD986:
Was fixed with 9fafd978a02b6e30d9044a68c15ea0e6e2061326. Thanks!

PR #294, opened 2016-01-28
Add decoding of L2C capability mask

dt-exafore:
At the moment discussion is here https://github.com/swift-nav/exafore_planning/issues/38

PR #301, opened 2016-02-11
Peregrine updates

cbeighley:
- Update Python requirements to specific version numbers
- Add support to cython `NavMsg` class for comparison and pickling
- Fix some `__cinit__` methods

### DIFF
--- a/python/swiftnav/pvt.pyx
+++ b/python/swiftnav/pvt.pyx
@@ -65,7 +65,7 @@ def calc_PVT_(nav_meas, disable_raim=False):
     nav_meas_[n] = (<NavigationMeasurement?>nav_meas[n])._thisptr
   dops_ = DOPS()
   soln = GNSSSolution()
-  cdef u8 ret = calc_PVT(n_used, nav_meas_, disable_raim, &soln._thisptr, &dops_._thisptr)
+  cdef s8 ret = calc_PVT(n_used, nav_meas_, disable_raim, &soln._thisptr, &dops_._thisptr)
   if not ret == 0:
     warnings.warn(_calc_pvt_codes[ret])
   free(nav_meas_)


python/swiftnav/nav_msg.pyx
```diff
@@ -40,3 +43,57 @@ cdef class NavMsg:
   def process_subframe(self, e):
```
dt-exafore: Please, take into account that In order to implement L2C capability monitor (please, see PR https://github.com/swift-nav/libswiftnav/pull/294/ still not merged) I updated process_subframe. Now it takes pointer to gps_l1ca_decoded_data_t as the 2nd parameter. So I guess tmp type should be changed. However, I believe this should be done after my PR be merged.

python/swiftnav/nav_msg.pyx
```diff
@@ -40,3 +43,59 @@ cdef class NavMsg:
...
+    http://docs.cython.org/src/userguide/special_methods.html.
+    """
+    if op == 2:
+      return self._equal(other)
+    elif op == 3:
+      return not self._equal(other)
+    else:
+      raise NotImplementedError
+
+  def _equal(self, other):
```
mookerji: Needs a test!

python/swiftnav/nav_msg.pyx
```diff
@@ -31,6 +31,9 @@ cdef class NavMsg:
   def from_dict(self, d):
     self._thisptr = d
 
+  def __reduce__(self):
```
mookerji: Needs a test!

python/swiftnav/track.pyx
```diff
@@ -236,19 +237,18 @@ cdef class AidedTrackingLoop:
...
-                  self._thisptr.code_bw,
-                  self._thisptr.code_zeta,
-                  self._thisptr.code_k,
-                  self._thisptr.carr_to_code,
-                  self._thisptr.carr_freq,
-                  self._thisptr.carr_bw,
-                  self._thisptr.carr_zeta,
-                  self._thisptr.carr_k,
-                  self._thisptr.carr_freq_b1)
+                  kwargs['loop_freq'],
```
mookerji: I might do something like `kwargs.pop('loop_freq', <some default or nothing>)` here?

python/swiftnav/track.pyx
```diff
@@ -396,10 +404,10 @@ cdef class CN0Estimator:
 
   def __cinit__(self, **kwargs):
     cn0_est_init(&self._thisptr,
-                 self._thisptr.bw,
-                 self._thisptr.cn0_0,
-                 self._thisptr.cutoff_freq,
-                 self._thisptr.loop_freq)
+                 kwargs['bw'],
```
mookerji: Same pop pop pop stuff.

python/swiftnav/track.pyx
```diff
@@ -396,10 +404,10 @@ cdef class CN0Estimator:
...
+                 kwargs['cn0_0'],
+                 kwargs['cutoff_freq'],
+                 kwargs['loop_freq'])
 
   def update(self, I, Q):
     """
```
mookerji: In general, I don't think the tracking stuff in libswiftnav python has tests, so adding some sanity tests would be great.

PR #303, opened 2016-02-15
Refactor signal_t to use a single enum for constellation and code

jacobmcnamee:
Struct `gnss_signal_t` has been changed to
```
typedef struct {
  u16 sat;
  enum code code;
} gnss_signal_t;
```
where the `code` enum specifies constellation, band, modulation type:
```
enum code {
  CODE_INVALID = -1,
  CODE_GPS_L1CA,
  CODE_GPS_L2CM,
  CODE_SBAS_L1CA,
  CODE_COUNT,
};
```
Also removes all indexing other than code-specific indexing from libswiftnav. Users of the library (e.g. piksi_firmware) can provide their own global and/or constellation-specific indexing such that only supported code types are included. For example, if GPS_L2CM is not supported by a particular hardware platform, then that platform can provide a global indexing function which skips GPS_L2CM signals.


include/libswiftnav/signal.h
```diff
@@ -15,68 +16,85 @@
...
+/** \addtogroup signal
+ * \{ */
+
+/* Number of satellites in each constellation. */
 #define NUM_SATS_GPS 32
 #define NUM_SATS_SBAS 19
 #define NUM_SATS (NUM_SATS_GPS + NUM_SATS_SBAS)
 
+/* Number of codes in each constellation. */
+#define NUM_CODES_GPS 2
```
pmiettinen: These are kind of redundant and never used if NUM_SIGNALS_<code> defines are used. Although I am a little bit iffy on those defines since is there ever a situation where they don't equal NUMS_SATS_<constellation> values?

src/signal.c
```diff
@@ -15,113 +16,133 @@
...
-} sat_group_t;
+/** \defgroup signal GNSS signal identifiers (SID)
+ * \{ */
 
-static const char * constellation_strs[CONSTELLATION_COUNT] = {
-  [CONSTELLATION_GPS] = "GPS",
-  [CONSTELLATION_SBAS] = "SBAS"
-};
+/** Macro used to assert that the specified code is valid. */
+#define ASSERT_CODE_VALID(code) assert(((code) >= 0) && ((code) < CODE_COUNT))
```
pmiettinen: not used anywhere

src/signal.c
```diff
@@ -15,113 +16,133 @@
...
-        .constellation = sat_groups[group_index].constellation,
-        .band = sat_groups[group_index].band,
-        .sat = sat_groups[group_index].sat_get(offset)
-      };
-      return sid;
-    }
-  }
-  assert("sid_from_index() failed");
-  return sid;
+  assert((code >= 0) && (code < CODE_COUNT));
```
pmiettinen: Use ASSERT_CODE_VALID macro?

src/signal.c
```diff
@@ -15,113 +16,133 @@
...
+
+/** Get the constellation to which a code belongs.
+ *
+ * \param code  Code to use.
+ *
+ * \return Constellation to which code belongs.
+ */
+enum constellation code_to_constellation(enum code code)
+{
+  assert((code >= 0) && (code < CODE_COUNT));
```
pmiettinen: Use ASSERT_CODE_VALID macro?

src/prns.c
```diff
@@ -14,8 +15,18 @@
...
+#define PRN_CODE_LENGTH_BYTES 128
+
+static const u8 gps_l1ca_codes[][PRN_CODE_LENGTH_BYTES];
+static const u8 sbas_l1ca_codes[][PRN_CODE_LENGTH_BYTES];
+
+/* Table of arrays of PRN codes indexed by code type */
+typedef const u8 (*prn_array_t)[PRN_CODE_LENGTH_BYTES];
+static const prn_array_t prn_array_table[CODE_COUNT] = {
+  [CODE_GPS_L1CA] = gps_l1ca_codes,
+  [CODE_GPS_L2CM] = 0,
```
pmiettinen: For consistency use NULL? (same as ca_code return value in error case)

src/prns.c
```diff
@@ -35,19 +46,13 @@ static const u8 sbas_ca_codes[][128];
...
-      assert(sid.sat - GPS_FIRST_PRN < NUM_SATS_GPS);
-      return gps_ca_codes[sid.sat - GPS_FIRST_PRN];
-    case CONSTELLATION_SBAS:
-      assert(sid.sat - SBAS_FIRST_PRN < NUM_SATS_SBAS);
-      return sbas_ca_codes[sid.sat - SBAS_FIRST_PRN];
-    default:
-      assert("unsupported constellation");
+  assert(sid_valid(sid));
+  const prn_array_t prn_array = prn_array_table[sid.code];
+  if (prn_array == 0) {
```
pmiettinen: NULL?

src/signal.c
```diff
@@ -15,113 +16,133 @@
...
+  enum constellation constellation;
+  u16 sat_count;
+  u16 sat_start;
+  const char *str;
+} code_table_element_t;
+
+/** Table of useful data for each code. */
+static const code_table_element_t code_table[CODE_COUNT] = {
+  [CODE_GPS_L1CA] =
+    {CONSTELLATION_GPS, NUM_SIGNALS_GPS_L1CA, GPS_FIRST_PRN, "GPS L1CA"},
```
pmiettinen: Variable is named 'sat_count' but the set value is number of signals. (see the comment in signal.h). Not even near fatal but a little bit incoherent.

include/libswiftnav/signal.h
jacobmcnamee: I think they could be useful if we need code-specific lookups for a single constellation. I did add a test to make sure the NUM_CODES_\<constellation\> defines are consistent with what code_to_constellation() returns.

Hopefully NUM_SIGNALS_\<code\> will always equal NUMS_SATS_\<constellation\>, but I do like the abstraction. If we ever add another field or something I could see it changing.

src/prns.c
jacobmcnamee: fixed

src/prns.c
jacobmcnamee: fixed

src/signal.c
jacobmcnamee: removed macro, added code_valid() and constellation_valid()

src/signal.c
jacobmcnamee: code_valid() function

src/signal.c
jacobmcnamee: code_valid() function

src/signal.c
jacobmcnamee: Yeah, agree. I do think it's better as is than changing it to signal_count and making the assumption that signal_count is equivalent to sat_count in the code.

PR #305, opened 2016-02-18
Improve dgnss python bindings

ljbade:
Use `mk_sdiff_array` in the `dgnss_update` binding.

Also add some `DEBUG` logging for the dgnss functions. Useful for debugging problems with the filter. Based on @akleeman work

@swift-nav/firmware Ready for review.

### DIFF
--- a/src/dgnss_management.c
+++ b/src/dgnss_management.c
@@ -250,24 +250,21 @@ void dgnss_update(u8 num_sats, sdiff_t *sdiffs, double receiver_ecef[3],
                   bool disable_raim, double raim_threshold)
 {
   DEBUG_ENTRY();
-  if (DEBUG) {
-    printf("dgnss_update\n");
-    printf("============\n");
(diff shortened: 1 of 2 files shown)


src/dgnss_management.c
```diff
@@ -251,11 +251,23 @@ void dgnss_update(u8 num_sats, sdiff_t *sdiffs, double receiver_ecef[3],
 {
   DEBUG_ENTRY();
   if (DEBUG) {
-    printf("sdiff[*].prn = {");
+    printf("dgnss_update\n");
```
mookerji: Note that you can use log_debug?

ljbade: Oh yeah that probably be a better choice!

ljbade:
@mookerji I replaced the `printf` with `log_debug`

mookerji:
Looks good, @ljbade ! Go ahead and merge whenever.

PR #308, opened 2016-02-22
WIP: Add necessary changes for L2C track loop

dt-exafore:
This PR contains necessary updates for https://github.com/swift-nav/piksi_firmware/pull/618 and should be merged before it..


dt-exafore:
@adel-mamin Please, review.

adel-mamin:
Looks good to me.

include/libswiftnav/constants.h
```diff
@@ -56,10 +59,6 @@
  * \note This is GPS_C / mu where mu is 1.0002926 */
 #define GPS_C_NO_VAC (GPS_C / 1.0002926)
 
-/** The wavelength of L1 in a vacuum.
- * \note This is GPS_C / GPS_L1_HZ. */
-#define GPS_L1_LAMBDA (GPS_C / GPS_L1_HZ)
```
ljbade: Why is this removed?

dt-exafore: This was removed to avoid implementing one more utility function like code_to_carr_freq. There is only place (observation.c) where the macro is used so I simple changed it with GPS_C/code_to_carr_freq(e[i]->sid.code) see below.

PR #315, opened 2016-03-21
Peregrine l2c update

adel-mamin:
Peregrine L2C related changes for libswiftnav


python/setup.py
```diff
@@ -24,19 +24,33 @@
...
+  library_dirs.append(os.path.expanduser('~/.local/lib'))
+
+  # Additional include directories:
+  # - Numpy includes
+  # - User local enviroment
+  # - Current directory
+  include_dirs = []
+  include_dirs.append(np.get_include())
+  include_dirs.append(os.path.expanduser('~/.local/include'))
+  include_dirs.append('.')
```
gsmcmullin: Having removed `../include/` from here causes the python build to fail on Travis where the libraries are installed to an non-standard location.  Maybe just have Travis install to `~/.local/`?

pmiettinen:
Superseded by https://github.com/swift-nav/libswiftnav/pull/350

PR #316, opened 2016-03-28
Additional validation of decoded ephemeris

ljbade:
Provides additional validation of decoded ephemeris by checking some values for valid ranges as per http://www.gps.gov/technical/icwg/meetings/2015/PIRN-IS-200H-003.pdf

Additional validation will also be needed in the firmware.


ljbade:
Should the validation occur in decoding, or in the `ephemeris_valid` function?

Advantage of checking vales in the valid function would detect errors received from external sources such as SBP or RINEX.

Would require a lot more checks for the other values to ensure they are also in range.

PR #317, opened 2016-03-28
remove ancient plover build test

kovach:
@gsmcmullin this should fix the build issue you and others have encountered

I will make a separate PR with an updated plover build script


scottswift:
closed in favor of #318

PR #319, opened 2016-03-31
Remove printf() from sid_to_string()

jacobmcnamee:
This reduces execution time on STM from 15us to 2us.


ljbade:
Is `ltoa()` not fast enough for the integer conversion?

Also where is this function being called often enough to be causing problems? This function should really only be needed for warnings and such

ljbade:
Nice! Looks good @jacobmcnamee

gsmcmullin:
lgtm

PR #321, opened 2016-04-04
Glo correlator

pmiettinen:
Adel's peregrine-l2c-update needs to be merged first.

https://github.com/swift-nav/libswiftnav/pull/315


valeri-atamaniouk:
Replaced with https://github.com/swift-nav/libswiftnav/pull/365

PR #320, opened 2016-03-31
Add GLO ephemeris

rge-exafore:
This commit adds definition of Glonass ephemeris data structure.

PR #323, opened 2016-04-06
WIP: Add GLO orbits computation

dt-exafore:
Refer https://github.com/swift-nav/exafore_planning/issues/106 to see dependencies.

### DIFF
--- a/include/libswiftnav/constants.h
+++ b/include/libswiftnav/constants.h
@@ -41,7 +41,7 @@
  * \note This is actually not identical to the usual WGS84 definition. */
 #define GPS_OMEGAE_DOT 7.2921151467e-5
 
-/** Earth’s Gravitational Constant as defined in the ICD in m^3 / s^2
+/** Earth's Gravitational Constant as defined in the ICD in m^3 / s^2
  * \note This is actually not identical to the usual WGS84 definition. */
 #define GPS_GM 3.986005e14
 
@@ -69,6 +69,22 @@
 
 /** GPS C/A code chipping rate in Hz. */
 #define GPS_CA_CHIPPING_RATE 1.023e6
+
+/** GLO semi-major axis of Earth
+  * NOTE: there is define WGS84_A which is 6378137, differ than defined
+  * in GLO ICD, refer A.3.1.2. */
+#define GLO_A_E 6378136.0
+
+/** Second zonal harmonic of the geopotential */
+#define GLO_J02 1.0826257e-3
+
+/** Earth's Gravitational Constant as defined in the GLO ICD in m^3 / s^2
+ * \note This is actually not identical to the usual WGS84 definition. */
+#define GLO_GM 3.9860044e14
+
+/** Earth's rotation rate as defined in the GLO ICD in rad / s
+ * \note This is actually not identical to the usual WGS84 definition. */
+#define GLO_OMEGAE_DOT 7.292115e-5
 
 /* \} */
 
--- a/python/tests/test_ephemeris.py
+++ b/python/tests/test_ephemeris.py
@@ -66,7 +66,7 @@ def test_sat_state():
   assert eph.is_valid(t.GpsTime(**{ 'wn': 1867, 'tow': 518400.0 + 3600.*1}))
   assert eph.is_valid(t.GpsTime(**{ 'wn': 1867, 'tow': 518400.0 - 3600.*1}))
   assert not eph.is_valid(t.GpsTime(**{ 'wn': 1868, 'tow': 518400.0,}))
-  assert len(eph.to_dict()) == 8
+  assert len(eph.to_dict()) == 9
   assert repr(eph)
   eph = e.Ephemeris(**{'sid': {'sat': 17, 'code': 0},
                      'toe': { 'wn': 1867, 'tow': 518400.0,},
@@ -193,3 +193,71 @@ def test_parameters():
   valid = 1
   fit_interval = 4 * 60 * 60
   assert not e.Ephemeris.is_params_valid(valid, fit_interval, toe, time)
+
+def test_glo_sat_state():
+  eph1 = e.Ephemeris(**{'sid': {'sat': 3, 'code': 3,},
+                     'toe': {'wn': 1892, 'tow': 303900.0,}, #from GPS
+                     'ura': 2.0,
+                     'fit_interval': 4 * 60 * 60,
+                     'valid': 1,
+                     'healthy': 1,
+                     'glo': {  #data at 2016-04-13T12:25:00
+                              'gamma' : 9.09494701772928238e-13,
+                              'tau' : -6.72144815325737000e-05,
+                              'pos' : [-1.4379786621093750e+07,-7.6587543945312500e+06,1.9676794921875000e+07],
+                              'vel' : [2.6736078262329102e+03,-2.4776077270507813e+02,1.8579368591308594e+03],
+                              'acc' : [0.0,0.0,-2.79396772384643555e-06],
+                            }})
+  eph2 = e.Ephemeris(**{'sid': {'sat': 3, 'code': 3,},
+                     'toe': {'wn': 1892, 'tow': 305700.0,}, #from GPS
+                     'ura': 2.0,
+                     'fit_interval': 4 * 60 * 60,
+                     'valid': 1,
+                     'healthy': 1,
+                     'glo': {  #data at 2016-04-13T12:55:00
+                              'gamma' : 9.09494701772928238e-13,
+                              'tau' : -6.72172755002975464e-05,
+                              'pos' : [-9.2793393554687500e+06,-8.5263706054687500e+06,2.2220868652343750e+07],
+                              'vel' : [2.9441843032836914e+03,-7.2328472137451172e+02,9.5054054260253906e+02],
+                              'acc' : [0.0,0.0,-2.79396772384643555e-06],
+                            }})
+  #check that both ephemeris are valid at 2016-04-13T12:40:00
+  assert eph1.is_valid(t.GpsTime(**{ 'wn': 1892, 'tow': 304800.0,}))
+  assert eph2.is_valid(t.GpsTime(**{ 'wn': 1892, 'tow': 304800.0,}))
+  assert eph1.is_healthy()
+  assert eph2.is_healthy()
+  #set time in between 2 ephemeris
+  time = t.GpsTime(**{ 'wn': 1892, 'tow': 304800.0,})
+  #calculate SV orbits at the time
+  pos1, vel1, clock_err1, clock_rate_err1 =  eph1.calc_sat_state(time)
+  pos2, vel2, clock_err2, clock_rate_err2 =  eph2.calc_sat_state(time)
+  #check if position difference of the SV in 2.5 m range
+  assert np.allclose(pos1, pos2, 0, 2.5)
+  #check if velocity difference of the SV in 0.005 m/s range
+  assert np.allclose(vel1, vel2, 0, 0.005)
+  assert abs(clock_err1 - clock_err2) < 1.0e-8
+  assert abs(clock_rate_err1 - clock_rate_err2) < 1.0e-8
+  # Now check GLO orbit computation (pos and vel) at particular time
+  # set reference ephemeris at 2007-15-11T06.15.00
+  eph = e.Ephemeris(**{'sid': {'sat': 3, 'code': 3,},
+                    'toe': {'wn': 1453, 'tow': 368100.0,}, #from GPS
+                    'ura': 2.0,
+                    'fit_interval': 4 * 60 * 60,
+                    'valid': 1,
+                    'healthy': 1,
+                    'glo': {  #input data from GLO ICD, pg. 55
+                             'gamma' : 0, #not necessary for the test
+                             'tau' : 0, #not necessary for the test
+                             'pos' : [-14081752.701,18358958.252,10861302.124],
+                             'vel' : [-1025.76358,1086.72147,-3157.32343],
+                             'acc' : [1.7156e-6-9.2581e-7,1.0278e-6-1.0343e-6,-1.0368e-6-1.1260e-6],
+                          }})
+  #set time to 6:30
+  time = t.GpsTime(**{ 'wn': 1453, 'tow': 369000.0,})
+  #calculate SV orbits at the time
+  pos, vel, clock_err, clock_rate_err =  eph.calc_sat_state(time)
+  #check calculated vel and pos data at 2007-15-11T06:30:00
+  #reference values from GLO ICD, pg 55. Note: it seems there is an typo in
+  # result Vz in ICD example
+  assert np.allclose(pos, [-14836563.872,19249935.476,7924017.196], 0, 2.5)
+  assert np.allclose(vel, [-653.97782,882.62958,-3359.44444], 0, 0.005)
--- a/src/ephemeris.c
+++ b/src/ephemeris.c
@@ -24,6 +24,8 @@
 
 float decode_ura_index(const u8 index);
 u32 decode_fit_interval(u8 fit_interval_flag, u16 iodc);
+/* maximum step length in seconds for Runge-Kutta aglorithm */
+#define GLO_MAX_STEP_LENGTH 30
 
 /** \defgroup ephemeris Ephemeris
  * Functions and calculations related to the GPS ephemeris.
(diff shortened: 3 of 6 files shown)


dt-exafore:
@nsirola, please review.

src/ephemeris.c
```diff
@@ -68,6 +70,129 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+              pos[0] * (1 - 5 * pow(pos[2], 2) / pow(r, 2))
+            + pow(GPS_OMEGAE_DOT, 2) * pos[0]
+            + 2 * GPS_OMEGAE_DOT * vel[1]
+            + acc[0];
+
+  ydot[4] = -GPS_GM / pow(r,3) * pos[1]
+            - 3/2 * GLO_J02 * GPS_GM * pow(GLO_A_E, 2) / pow(r, 5) *
+              pos[1] * (1 - 5 * pow(pos[2], 2) / pow(r, 2))
+            + pow(GPS_OMEGAE_DOT, 2) * pos[1]
+            + 2 * GPS_OMEGAE_DOT * vel[0]
```
ljbade: This should be -2

src/ephemeris.c
```diff
@@ -68,6 +70,129 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+
+  double r = sqrt(pow(pos[0], 2) +
+                  pow(pos[1], 2) +
+                  pow(pos[2], 2));
+
+  ydot[0] = vel[0];
+  ydot[1] = vel[1];
+  ydot[2] = vel[2];
+
+  ydot[3] = -GPS_GM / pow(r,3) * pos[0]
```
ljbade: It is possible to simplify common terms between the equations - see https://github.com/swift-nav/gnss-analysis/blob/master/gnss_analysis/ephemeris.py#L409

src/ephemeris.c
```diff
@@ -68,6 +70,129 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+            - 3/2 * GLO_J02 * GPS_GM * pow(GLO_A_E, 2) / pow(r, 5) *
```
ljbade: all double constants should be expressed with the decimal, e.g. `3.0 / 2.0`

ljbade: Possible simplification for terms that are `x / pow(r ,5)` would be `x / pow(r, 1.0 / 5.0)`

src/ephemeris.c
ljbade: You can't use `GPS_GM` you will need to add a new `GLO_GM` with value 3.9860044e14

src/ephemeris.c
```diff
@@ -68,6 +70,129 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+              pos[0] * (1 - 5 * pow(pos[2], 2) / pow(r, 2))
+            + pow(GPS_OMEGAE_DOT, 2) * pos[0]
```
ljbade: You can't use `GPS_OMEGAE_DOT` you will need `GLO_OMEGAE_DOT` with value 7.292115e-5

src/ephemeris.c
```diff
@@ -68,6 +70,129 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+  if (ceil(dt / GLO_NUM_STEPS)) {
+    double ydot0[6], yn[6];
+    double h = dt / GLO_NUM_STEPS;
+
+    calc_ydot(ydot0, e->glo.pos, e->glo.vel, e->glo.acc);
+    memcpy(&yn[0], e->glo.pos, sizeof(double) * 3);
+    memcpy(&yn[3], e->glo.vel, sizeof(double) * 3);
+
+    /* Runge-Kutta integration algorithm */
+    for (u8 i = 0; i < GLO_NUM_STEPS; i++) {
```
ljbade: There appears to be no control of the step size? RTCM recommends maximum step size of 30 sec.

src/ephemeris.c
```diff
@@ -68,6 +70,129 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+
+      calc_ydot(k3, &y_tmp[0], &y_tmp[3], e->glo.acc);
+
+      for (j = 0; j < 6; j++)
+        y_tmp[j] = yn[j] + h * k3[j];
+
+      calc_ydot(k4, &y_tmp[0], &y_tmp[3], e->glo.acc);
+
+      for (j = 0; j < 6; j++)
+        yn[j] += h/6 * (k1[j] + k2[j] + k3[j] + k4[j]);
```
ljbade: This wrong, should be k1 + 2 * k2 + 2* k3 + k4

src/ephemeris.c
```diff
@@ -68,6 +70,129 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+      calc_ydot(ydot0, &yn[0], &yn[3], e->glo.acc);
+    }
+    memcpy(pos, &yn[0], sizeof(double) * 3);
+    memcpy(vel, &yn[3], sizeof(double) * 3);
+  } else {
+    memcpy(pos, e->glo.pos, sizeof(double) * 3);
+    memcpy(vel, e->glo.vel, sizeof(double) * 3);
+  }
+
+  *clock_err = e->glo.tau;
```
ljbade: This is wrong, see https://github.com/tomojitakasu/RTKLIB/blob/master/src/ephemeris.c#L288

src/ephemeris.c
```diff
@@ -68,6 +70,129 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+  *clock_rate_err = e->glo.gamma;
```
ljbade: Same as ^

include/libswiftnav/signal.h
```diff
@@ -22,24 +22,30 @@
...
 
 /* Number of signals in each constellation. */
 #define NUM_SIGNALS_GPS (NUM_SIGNALS_GPS_L1CA + NUM_SIGNALS_GPS_L2CM)
 #define NUM_SIGNALS_SBAS (NUM_SIGNALS_SBAS_L1CA)
 #define NUM_SIGNALS (NUM_SIGNALS_GPS + NUM_SIGNALS_SBAS)
+#define NUM_SIGNALS_GLO (NUM_SIGNALS_GLO_L1CA + NUM_SIGNALS_GLO_L2CM)
 
 #define GPS_FIRST_PRN 1
 #define SBAS_FIRST_PRN 120
+#define GLO_FIRST_PRN 64 //NOTE: DT: check this
```
ljbade: Based on question to @jacobmcnamee, this should be 1

ljbade:
@dt-exafore I found a number of problems and commented them

src/ephemeris.c
dt-exafore: Do you mean line 107? What about line 100?

ljbade: Yeah so instead of ` + 2 * GPS_OMEGAE_DOT * vel[0]` it should be ` - 2 * GPS_OMEGAE_DOT * vel[0]`

If you look at the RTCM document or the Matlab code you will see line 100 should be +2 coefficient, and 107 a -2 coeffcient

src/ephemeris.c
dt-exafore: I'll learn from @nsirola

include/libswiftnav/constants.h
```diff
@@ -70,6 +70,22 @@
...
+  * NOTE: there is define WGS84_A which is 6378137, differ than defined
+  * in GLO ICD, refer A.3.1.2. */
+#define GLO_A_E 6378136.0
+
+/** Second zonal harmonic of the geopotential */
+#define GLO_J02 1.0826257e-15
+
+/** Earth�s Gravitational Constant as defined in the GLO ICD in m^3 / s^2
+ * \note This is actually not identical to the usual WGS84 definition. */
+#define GLO_GM 3.9860044e14
```
ljbade: I think I might have been wrong. Older GLONASS ICD (using PZ-90.02) use this value. But it appears the latest ICD and http://www.unoosa.org/pdf/icg/2014/wg/PZ-90.11_v.1.1_01.11.2015.pdf use `3.986004418e14`.

It would have a very small (possibly not noticeable?) effect on the orbit result.

I wonder if there is a way to verify which value to use.

include/libswiftnav/signal.h
```diff
@@ -22,24 +22,30 @@
 /* Number of satellites in each constellation. */
 #define NUM_SATS_GPS 32
 #define NUM_SATS_SBAS 19
-#define NUM_SATS (NUM_SATS_GPS + NUM_SATS_SBAS)
+#define NUM_SATS_GLO 24 //NOTE: DT: check this
```
ljbade: This should be correct. Current GLONASS ICD only supports up to 24 satellites. However there are plans to expand the GLONASS almanac to support more satellites in the future.

src/ephemeris.c
ljbade: Actually now I am not sure. See http://www.unoosa.org/pdf/icg/2014/GLONASSTime.pdf

I wonder if there is someone who is expert in GLONASS to verify the correct calculation method.

ljbade: I checked another open source implementation of the clock correction in GPSTk, and they don't do iterations.

Basically the equation should be:
`*clock_err = e->glo.tau + e->glo.gamma * dt;`

include/libswiftnav/constants.h
nsirola: RTCM refers to ICD appendix A.3.1.2 so I used the value found there (398600.44).
Anyway the difference results in less than 0,0002 m change in computed track over 15 minutes.

src/ephemeris.c
```diff
@@ -21,6 +21,8 @@
 #include <libswiftnav/constants.h>
 #include <libswiftnav/ephemeris.h>
 
+#define GLO_NUM_STEPS 30 /* For Runge-Kutta algorithm */
```
nsirola: this should be 
/* maximum step length in seconds for Runge-Kutta aglorithm */
#define GLO_MAX_STEP_LENGTH 30

src/ephemeris.c
nsirola: Instead of constant number of steps, we should have constant maximum step size (see earlier comment), and the number of steps and actual step size (which will always be less) computed here as:

num_steps = ceil(dt / GLO_MAX_STEP_LENGTH);
h = dt / num_steps;

src/ephemeris.c
nsirola: Agreed.

src/ephemeris.c
```diff
@@ -68,6 +70,136 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+
+  double dt = abs(gpsdifftime(t, &e->toe));
+
+  if (dt > 900) {
+    log_error("GLO: Integration end point is not within 900 s of TOE");
+    return 1;
+  }
+
+  if (ceil(dt / GLO_NUM_STEPS)) {
+    double ydot0[6], yn[6];
```
nsirola: It would make the algorithm clearer if you renamed these variables to ydot and y.
(the 0 usually refers to the initial value and n to the final value, which is not how these variables are used here)

src/ephemeris.c
```diff
@@ -152,16 +153,18 @@ static s8 calc_sat_state_glo(const ephemeris_t *e, const gps_time_t *t,
     return 1;
   }
 
-  if (ceil(dt / GLO_NUM_STEPS)) {
+  double num_steps = ceil(dt / GLO_MAX_STEP_LENGTH);
```
nsirola: should be integer

src/ephemeris.c
```diff
@@ -152,16 +153,18 @@ static s8 calc_sat_state_glo(const ephemeris_t *e, const gps_time_t *t,
...
     double ydot0[6], yn[6];
-    double h = dt / GLO_NUM_STEPS;
 
     calc_ydot(ydot0, e->glo.pos, e->glo.vel, e->glo.acc);
     memcpy(&yn[0], e->glo.pos, sizeof(double) * 3);
     memcpy(&yn[3], e->glo.vel, sizeof(double) * 3);
 
     /* Runge-Kutta integration algorithm */
-    for (u8 i = 0; i < GLO_NUM_STEPS; i++) {
+    for (u8 i = 0; i < GLO_MAX_STEP_LENGTH; i++) {
```
nsirola: should be "i < num_steps"

ljbade:
Looks like @nsirola last comments got hidden for some reason

src/ephemeris.c
```diff
@@ -68,6 +71,138 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+  assert(clock_rate_err != NULL);
+
+  double dt = abs(gpsdifftime(t, &e->toe));
+
+  if (dt > 900) {
+    log_error("GLO: Integration end point is not within 900 s of TOE");
+    return 1;
+  }
+
+  double num_steps = ceil(dt / GLO_MAX_STEP_LENGTH);
```
nsirola: Should be integer
(trying again, this time commenting on the current code)

src/ephemeris.c
```diff
@@ -68,6 +71,138 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+
+  if (num_steps) {
+    double ydot0[6], yn[6];
+
+    calc_ydot(ydot0, e->glo.pos, e->glo.vel, e->glo.acc);
+    memcpy(&yn[0], e->glo.pos, sizeof(double) * 3);
+    memcpy(&yn[3], e->glo.vel, sizeof(double) * 3);
+
+    /* Runge-Kutta integration algorithm */
+    for (u8 i = 0; i < GLO_MAX_STEP_LENGTH; i++) {
```
nsirola: should be "i < num_steps"

ljbade:
@nsirola I think you also had one about the variable names

src/ephemeris.c
```diff
@@ -68,6 +71,138 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+  if (dt > 900) {
+    log_error("GLO: Integration end point is not within 900 s of TOE");
+    return 1;
+  }
+
+  double num_steps = ceil(dt / GLO_MAX_STEP_LENGTH);
+  double h = dt / num_steps;
+
+  if (num_steps) {
+    double ydot0[6], yn[6];
```
nsirola: It would make the algorithm clearer if you renamed these variables to ydot and y.
(the 0 usually refers to the initial value and n to the final value, which is not how these variables are used here)

src/ephemeris.c
```diff
@@ -68,6 +71,138 @@ static s8 calc_sat_state_xyz(const ephemeris_t *e, const gps_time_t *t,
...
+
+  double dt = abs(gpsdifftime(t, &e->toe));
+
+  if (dt > 900) {
+    log_error("GLO: Integration end point is not within 900 s of TOE");
+    return 1;
+  }
+
+  double num_steps = ceil(dt / GLO_MAX_STEP_LENGTH);
+  double h = dt / num_steps;
```
nsirola: to preserve the correct direction of step, correct to
double h = gpsdifftime(t,&e->toe) / num_steps;

include/libswiftnav/constants.h
```diff
@@ -70,6 +70,22 @@
 /** GPS C/A code chipping rate in Hz. */
 #define GPS_CA_CHIPPING_RATE 1.023e6
 
+/** GLO semi-major axis of Earth
+  * NOTE: there is define WGS84_A which is 6378137, differ than defined
+  * in GLO ICD, refer A.3.1.2. */
+#define GLO_A_E 6378136.0
+
+/** Second zonal harmonic of the geopotential */
+#define GLO_J02 1.0826257e-15
```
nsirola: Should be
1082625.7e-9 (as per ICD A.3.1.2)

src/ephemeris.c
dt-exafore: corrected

python/tests/test_ephemeris.py
```diff
@@ -186,3 +186,48 @@ def test_parameters():
...
+
+def test_glo_sat_state():
+  eph1 = e.Ephemeris(**{'sid': {'sat': 1, 'code': 3,},
+                     'toe': {'wn': 1884, 'tow': 260100.0,}, #from GPS
+                     'ura': 2.0,
+                     'fit_interval': 4 * 60 * 60,
+                     'valid': 1,
+                     'healthy': 1,
+                     'glo': {  #RINEX data at 2016-02-17T00:15:00
+                              'gamma' : 0,
```
ljbade: Hmm are there any satellites with non-zero gamma? Otherwise this wouldn't pick up an error in using gamma to compute the clock correction.

python/tests/test_ephemeris.py
```diff
@@ -186,3 +186,48 @@ def test_parameters():
...
+  #calculate SV orbits at the time
+  pos1, vel1, clock_err1, clock_rate_err1 =  eph1.calc_sat_state(time)
+  pos2, vel2, clock_err2, clock_rate_err2 =  eph2.calc_sat_state(time)
+  clock_rate_err1-clock_rate_err2
+  #check if position difference of the SV in 2.5 m range
+  assert np.allclose(pos1, pos2, 0, 2.5)
+  #check if velosity difference of the SV in 0.005 m/s range
+  assert np.allclose(vel1, vel2, 0, 0.005)
+  assert abs(clock_err1 - clock_err2) < 1.0e-9
+  assert abs(clock_rate_err1 - clock_rate_err2) < 1.0e-9
```
ljbade: Id also add one test that just checks the computation against a fixed "correct" satellite position similar to what is in the GPS test.

src/ephemeris.c
nsirola: Sorry, meant num_steps should be integer.
h has to be double.
(which reminds me, need to have some test with non-integer t resulting in non-integer step-size)

dt-exafore: right, my bad

python/tests/test_ephemeris.py
dt-exafore: I'm not sure I follow. Do you mean test case like this: calculate position nd velocity for, say, 15 mins off from received ephemeris and compare the results with data that should be at the time? Like lines 52 and 53?

ljbade: Yep, you could just add something like line 52/53 to this test

python/tests/test_ephemeris.py
dt-exafore: Yes, they are. I was able to collect GLO ephemeris. I'll update the tests.

python/tests/test_ephemeris.py
dt-exafore: Could you please clarify how reference position and velocity were calculated at lines 52-53? If I use real data from ephemeris I'll have huge difference, because ephemeris keeps unchanged within 30 mins.

dt-exafore: done

python/tests/test_ephemeris.py
dt-exafore: done

nsirola:
Looks good to me

ljbade:
@dt-exafore Can you rebase on latest master? It should pull in @rge-exafore's final GLONASS defines in `signal.h`

dt-exafore:
Rebased, to master, but since PR https://github.com/swift-nav/libswiftnav/pull/320 is still not merged I have build errors.

dt-exafore:
rebased to master tip.
PR #320 still needed

include/libswiftnav/constants.h
```diff
@@ -41,7 +41,7 @@
  * \note This is actually not identical to the usual WGS84 definition. */
 #define GPS_OMEGAE_DOT 7.2921151467e-5
 
-/** Earth’s Gravitational Constant as defined in the ICD in m^3 / s^2
+/** Earth�s Gravitational Constant as defined in the ICD in m^3 / s^2
```
ljbade: can the `’` be replaced with `'`

include/libswiftnav/constants.h
```diff
@@ -70,6 +70,22 @@
...
 
+/** GLO semi-major axis of Earth
+  * NOTE: there is define WGS84_A which is 6378137, differ than defined
+  * in GLO ICD, refer A.3.1.2. */
+#define GLO_A_E 6378136.0
+
+/** Second zonal harmonic of the geopotential */
+#define GLO_J02 1.0826257e-3
+
+/** Earth�s Gravitational Constant as defined in the GLO ICD in m^3 / s^2
```
ljbade: can the `’` be replaced with `'`

ljbade:
@dt-exafore Can you fix Travis?

include/libswiftnav/constants.h
dt-exafore: done

include/libswiftnav/constants.h
dt-exafore: done

ljbade:
Looks good. Merging.

PR #330, opened 2016-04-13
Add transformation of GLONASS-M current data information into gps_time_t

pmiettinen:


include/libswiftnav/time.h
```diff
@@ -17,6 +17,34 @@
...
+#define MINUTE_SECS         60
+
+/** Number of minutes in an hour. */
+#define HOUR_MINUTES        60
+
+/** Number of seconds in an hour. */
+#define HOUR_SECS           (MINUTE_SECS * HOUR_MINUTES)
+
 /** Number of seconds in a week. */
 #define WEEK_SECS (7*24*60*60)
```
ljbade: Perhaps we can also add an DAY_HOURS? then we could make this WEEK_DAYS * DAY_HOURS * HOUR_SECS

src/time.c
```diff
@@ -115,4 +117,72 @@ u16 gps_adjust_week_cycle(u16 wn_raw, u16 wn_ref)
...
+
+  /* Calculate days since GPS epoch */
+  s64 days_gps_epoch = YEAR_1980_GPS_DAYS + day_of_year - 1;
+  u32 y = GPS_EPOCH_YEAR;
+  while (++y < glo_year) {
+    days_gps_epoch += is_leap_year(y) ? LEAP_YEAR_DAYS : YEAR_DAYS;
+  }
+
+  gps_t.wn = days_gps_epoch / WEEK_DAYS;
+  gps_t.tow = GPS_MINUS_UTC_SECS + (days_gps_epoch % WEEK_DAYS) * DAY_SECS +
```
ljbade: Hmm this will break after the next leap second change...

I wonder how we can work around this?

include/libswiftnav/time.h
pmiettinen: Sure

src/time.c
pmiettinen: It will be a separate task to add proper leap second handling. This is how @rge-exafore instructed me.

pmiettinen:
There seems to be some problem with j = 3 and j = 4 cases, need to investigate.

pmiettinen:
@ljbade Please review

ljbade:
@pmiettinen Looks good for merging.

@pmiettinen Can you open an issue to mention that current GLONASS code uses fixed leap second value? And that we will need to implement proper GLONASS leap second handling before release of GLONASS support?

pmiettinen:
@ljbade https://github.com/swift-nav/exafore_planning/issues/167

ljbade:
@pmiettinen Excellent. Merging.

PR #334, opened 2016-04-21
Final CORS fix PR

ljbade:
Squashing into two commits to save time and get it ready for @denniszollo to start testing it ASAP.

### DIFF
--- a/include/libswiftnav/track.h
+++ b/include/libswiftnav/track.h
@@ -161,6 +161,7 @@ typedef struct {
 typedef struct {
   double raw_pseudorange;
   double pseudorange;
+  double raw_carrier_phase;
   double carrier_phase;
   double raw_doppler;
   double doppler;
@@ -234,12 +235,12 @@ void cn0_est_init(cn0_est_state_t *s, float bw, float cn0_0,
 float cn0_est(cn0_est_state_t *s, float I, float Q);
 
 s8 calc_navigation_measurement(u8 n_channels, const channel_measurement_t *meas[],
-                               navigation_measurement_t *nav_meas[], gps_time_t *nav_time,
+                               navigation_measurement_t *nav_meas[], gps_time_t *rec_time,
                                const ephemeris_t* e[]);
 
 int nav_meas_cmp(const void *a, const void *b);
 u8 tdcp_doppler(u8 n_new, navigation_measurement_t *m_new,
                 u8 n_old, navigation_measurement_t *m_old,
-                navigation_measurement_t *m_corrected);
+                navigation_measurement_t *m_corrected, double dt);
 
 #endif /* LIBSWIFTNAV_TRACK_H */
--- a/include/libswiftnav/track.h
+++ b/include/libswiftnav/track.h
@@ -234,7 +234,7 @@ void cn0_est_init(cn0_est_state_t *s, float bw, float cn0_0,
 float cn0_est(cn0_est_state_t *s, float I, float Q);
 
 s8 calc_navigation_measurement(u8 n_channels, const channel_measurement_t *meas[],
-                               navigation_measurement_t *nav_meas[],
+                               navigation_measurement_t *nav_meas[], gps_time_t *nav_time,
                                const ephemeris_t* e[]);
 
 int nav_meas_cmp(const void *a, const void *b);
--- a/src/baseline.c
+++ b/src/baseline.c
@@ -409,8 +409,9 @@ static bool chi_test(double threshold, u8 num_dds,
  *
  *   -`-1`: < 3 dds
  *   -`-2`: dgelsy  error (see lesq_solution_float)
- *   -`-3`: raim check failed, repair failed
+ *   -`-3`: raim check failed, repair failed, ref satellite was bad
  *   -`-4`: raim check failed, not enough sats for repair
+ *   -`-5`: raim check failed, repair failed, more than one acceptable solution
  */
 /* TODO(dsk) update all call sites to use n_used as calculated here.
  * TODO(dsk) add warn/info logging to call sites when repair occurs.
--- a/src/baseline.c
+++ b/src/baseline.c
@@ -501,7 +501,7 @@ s8 lesq_solve_raim(u8 num_dds_u8, const double *dd_obs,
     if (n_used) {
       *n_used = 0;
     }
-    return -3;
+    return -5;
   }
 }
 
--- a/src/track.c
+++ b/src/track.c
@@ -759,35 +759,54 @@ float cn0_est(cn0_est_state_t *s, float I, float Q)
   return s->log_bw - 10.f*log10f(s->nsr);
 }
 
-/**
+/** Calculate observations from tracking channel measurements.
+ *
+ * This function takes an array of pointers, for a version taking a flat array
+ * see calc_navigation_measurement().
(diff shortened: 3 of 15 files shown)


src/baseline.c
```diff
@@ -501,7 +501,7 @@ s8 lesq_solve_raim(u8 num_dds_u8, const double *dd_obs,
     if (n_used) {
       *n_used = 0;
     }
-    return -3;
+    return -5;
```
ejconlon: What does -5 mean? (add to docs)

src/track.c
```diff
@@ -760,50 +760,93 @@ float cn0_est(cn0_est_state_t *s, float I, float Q)
...
-    TOTs[i] += (nav_time - meas[i]->receiver_time) * meas[i]->code_phase_rate / 1.023e6;
+    nav_meas[i]->sid = meas[i]->sid;
 
-    /** \todo Maybe keep track of week number in tracking channel
-        state or derive it from system time. */
-    nav_meas[i]->tot.tow = TOTs[i];
+    /* Compute the time of transmit of the signal on the satellite from the
+     * tracking loop parameters. This will be used to compute the pseudorange. */
+    nav_meas[i]->tot.tow = 1e-3 * meas[i]->time_of_week_ms;
+    nav_meas[i]->tot.tow += meas[i]->code_phase_chips / 1.023e6;
```
jacobmcnamee: magic number

src/track.c
```diff
@@ -760,50 +760,93 @@ float cn0_est(cn0_est_state_t *s, float I, float Q)
...
+    nav_meas[i]->tot.tow += (rec_tc - meas[i]->receiver_time)
+                            * meas[i]->code_phase_rate / 1.023e6;
+    /* For now use the week number from the ephemeris. */
+    /* TODO: Should we use a more reliable source for the week number? */
+    /* TODO (Leith): There might be a bug where ephmeris ToW is set to 0 */
+    /* near end of the week, without ToW being rolled over? */
+    /* It woukd this functions' assumptions to break. */
```
jacobmcnamee: "It woukd this functions' assumptions to break" ??

src/track.c
```diff
@@ -828,11 +871,12 @@ int nav_meas_cmp(const void *a, const void *b)
  * \param n_old Number of measurements in `m_old`
  * \param m_old Array of old navigation measurements, sorted by PRN
  * \param m_corrected Array in which to store the output measurements
+ * \param dt The difference in receiver time betweeb the two measurements
```
jacobmcnamee: typo

src/track.c
```diff
@@ -760,50 +760,93 @@ float cn0_est(cn0_est_state_t *s, float I, float Q)
...
-    if ((TOTs[i] + clock_err[i]) > min_TOF)
-      min_TOF = TOTs[i];
+  /* To calculate the pseudorange from the time of transmit we need the local
+   * time of reception. */
+  gps_time_t tor;
+  if (rec_time) {
+    /* If we were given a time, use that. */
+    tor = *rec_time;
+  } else {
+    /* If we were not given a recieve time then we can just set one of the
```
jacobmcnamee: receive

jacobmcnamee:
trivial comments, otherwise lgtm

src/baseline.c
ljbade: Sure, I just changed this as there are currently two places that return -3!

src/track.c
ljbade: Is this now a #define per HW revision?

src/track.c
ljbade: lol

src/track.c
ljbade: Ah found `GPS_CA_CHIPPING_RATE`

ljbade:
@jacobmcnamee @ejconlon I addressed your comments.

src/baseline.c
jacobmcnamee: Just remove this from the PR. You'll need to update all the functions which call this one.

include/libswiftnav/track.h
```diff
@@ -146,8 +146,9 @@ typedef struct {
   double carrier_freq;     /**< Carrier frequency in Hz. */
   u32 time_of_week_ms;     /**< Number of milliseconds since the start of the
                                 GPS week corresponding to the last code rollover.  */
-  double receiver_time;    /**< Receiver clock time at which this measurement
-                                is valid in seconds. */
+  double rec_time_delta;   /**< Difference between receiver clock time at which
```
scottswift: hey, to fix the current travis build failure, I think you just need to carry over this update here:
https://github.com/swift-nav/libswiftnav/blob/master/python/swiftnav/track.pxd#L166

ljbade:
@scottswift Oh nice, wierd that Travis failed, must have been artefact from the rebase.

src/baseline.c
ljbade: We really do need to fix this return, is it silly to have to returns with the same value.

I plan to make another PR that adds error message strings for these codes, similar to `calc_PVT`

ljbade:
This PR now depends on https://github.com/swift-nav/libswiftnav/pull/347

ljbade:
Merging

src/track.c
```diff
@@ -796,19 +811,39 @@ s8 calc_navigation_measurement(u8 n_channels, const channel_measurement_t *meas[
...
-
+    /* The raw pseudorange is just the time of flight divided by the speed of
+     * light. */
+    nav_meas[i]->raw_pseudorange = GPS_C * gpsdifftime(&tor, &nav_meas[i]->tot);
+
+    /* The corrected pseudorange, carrier_phase, Doppler applies the clock error
+     * and clock rate error correction from the ephemeris respectively. */
+    nav_meas[i]->pseudorange = nav_meas[i]->raw_pseudorange
+                               + clock_err[i] * GPS_C;
+    nav_meas[i]->carrier_phase = nav_meas[i]->raw_carrier_phase
```
akleeman: I'm trying to figure this out, what does it mean to apply satellite clock adjustments to the carrier phase?

ljbade: Both the carrier frequency and data are effected by satellite clock error (they are generated from the same clock).

Thats my understanding,
@anth-cole Is this correct?

ljbade: Both the carrier frequency and data are affected by satellite clock error (they are generated from the same clock).

Thats my understanding,
@anth-cole Is this correct?

akleeman: The part that got me wondering if it's correct is that pseudorange uses the time of transmission (which includes some satellite clock error) in its derivation.  carrier phase however is just a relative thing that is integrated, without use of any reference time (to the best of my knowledge).

anth-cole: I think you do want to apply this correction to both. My understanding is that the phase is aligned to the ms boundaries on the code message. So this clock offset would end up being a part wavelength bias on the phase obs and so by removing it we move closer to the integer wavelength. Although, on the double difference level, this should cancel anyway so I would expect the same results if we apply it or not, it will become more significant in low-latency mode.

ljbade: @anth-cole yes currently the clock corrected pseudorange and carrier phase are only used for SPP

and the nav message and carrier phase wave are synced (they are both multiples of the fundamental GPS clock frequency)

PR #336, opened 2016-04-22
RTKLIB routines use Fortran storage order.

peddie:
@mookerji


peddie:
Tests are ready and improved if @mookerji @benjamin0 like it.

mookerji:
@peddie having checked that the expected values are just permuted from their expected locations, this looks totally good to me! 👍

PR #339, opened 2016-04-26
One instance of 'abs' should be 'fabs'

ejconlon:
clang on OSX complains about this.  -Werror fails the build.


ljbade:
Looks good, but can you look into adding osx job to Travis? Should ensure this doesn't happen again

ejconlon:
I'll submit this first and look into it.

PR #337, opened 2016-04-26
I123 glo decoder error detector

dt-exafore:
The commit adds error detection algorithm for GLO nav messages + unit tests.
Requires: https://github.com/swift-nav/libswiftnav/pull/329

PR #340, opened 2016-04-27
I181 glo decoder python binding

valeri-atamaniouk:
Python bindings for GLONASS message decoder

To be merged after #337


valeri-atamaniouk:
@gsmcmullin Please review. All dependencies are now satisfied.

python/setup.py
```diff
@@ -2,28 +2,30 @@
...
-  version = get_git_version(),
-  description = 'Python bindings to the libswiftnav library.',
-  license = 'LGPLv3',
-  url = 'http://www.swiftnav.com',
-  author = 'Swift Navigation Inc.',
-  author_email = 'dev@swiftnav.com',
-  maintainer = 'Swift Navigation',
-  maintainer_email = 'dev@swiftnav.com',
-  packages = ['swiftnav'],
+    name='swiftnav',
```
ljbade: Why is there extra spaces here?

python/setup.py
```diff
@@ -32,40 +34,41 @@ def make_extension(ext_name):
     if 'LD_LIBRARY_PATH' in os.environ:
       library_dirs.append(os.environ['LD_LIBRARY_PATH'])
     return Extension(
-      ext_name, [ext_path],
-      include_dirs = [np.get_include(), '.', '../include/'],
-      extra_compile_args = ['-O0', '-g'],
-      extra_link_args = ['-g'],
-      libraries = ['m', 'swiftnav'],
-      library_dirs = library_dirs,
+        ext_name, [ext_path],
```
ljbade: more extra spaces

python/setup.py
```diff
@@ -32,40 +34,41 @@ def make_extension(ext_name):
...
-    'swiftnav.baseline',
-    'swiftnav.bits',
-    'swiftnav.filter_utils',
-    'swiftnav.memory_pool',
-    'swiftnav.prns',
-    'swiftnav.sats_management',
-    'swiftnav.tropo',
-    'swiftnav.set',
-    'swiftnav.bit_sync',
+      'swiftnav.edc',
```
ljbade: same

python/setup.py
valeri-atamaniouk: The formatting has changed to to auto-formatting (autopep.8 --indent-size 2). Shall it be 2 here and below? Do we have any other non-default settings (other than indentation size)?

ljbade: Hmm, I will ask... @benjamin0 Do you know what the story is with our Python formatting?

benjamin0: @ljbade @valeri-atamaniouk https://swiftnav.hackpad.com/Engineering-Playbook-fKcJhOl85sD#:h=Python

ljbade: @valeri-atamaniouk looks like it should be two spaced ident

valeri-atamaniouk: @ljbade @benjamin0 https://www.python.org/dev/peps/pep-0008/#id3
https://www.python.org/dev/peps/pep-0008/#fn-hi
_-Hanging indents should add a level_

The issue is related to PEP8 conformance of E121 and E126. I suggest to update the coding style to mention that hanging indents are 2 spaces. It would make it impossible to use auto-indentation, but I'm perfectly fine with that.

For example, when using pep8 to check the style with only 2 spaces in hanging indents:

    $ pep8 --ignore E111,E114 setup.py
    setup.py:5:3: E121 continuation line under-indented for hanging indent
    setup.py:55:7: E121 continuation line under-indented for hanging indent
    setup.py:63:5: E121 continuation line under-indented for hanging indent

benjamin0: I think if `hanging indents should add a level` and we define a level as indented two spaces, then hanging indents are an additional 2 spaces. @valeri-atamaniouk feel free to update the linked hackpad above. @fnoble do you have an opinion?

PR #324, opened 2016-04-07
CN0 filter redesign + review comment fixes

Paakkit:
This PR supersedes https://github.com/swift-nav/libswiftnav/pull/322
This PR is supposed to be merged before https://github.com/swift-nav/piksi_firmware/pull/637

PR #341, opened 2016-05-04
I204 l1ca cn0 separation

valeri-atamaniouk:
CN0 processing functionality updated:
- Separated CN0 computation and value filtering
- Added SNV2 CN0 algorithm
- Added back LP1 filtering (original) in addition to BW2
- Added CN0 scaling according to sampling frequency

To be merged after #324


valeri-atamaniouk:
Too old.

PR #342, opened 2016-05-06
Fix time normalization function.

rge-exafore:
@gsmcmullin please review


ljbade:
Nice catch!

Are there unit tests for this function?

PR #345, opened 2016-05-12
Add calc_sat_state error code handling

rge-exafore:


ljbade:
Can you add a Doxygen comment for this function? It should also specify what the return codes mean, and what causes an error condition.

src/ephemeris.c
```diff
@@ -356,7 +356,11 @@ s8 calc_sat_state(const ephemeris_t *e, const gps_time_t *t,
 
   if (!ephemeris_valid(e, t)) {
     log_error_sid(e->sid,
-                  "Using invalid or too old ephemeris in calc_sat_state");
```
jacobmcnamee: Great.

src/track.c
```diff
@@ -760,9 +760,9 @@ float cn0_est(cn0_est_state_t *s, float I, float Q)
   return s->log_bw - 10.f*log10f(s->nsr);
 }
 
-void calc_navigation_measurement(u8 n_channels, const channel_measurement_t *meas[],
-                                 navigation_measurement_t *nav_meas[],
-                                 double nav_time, const ephemeris_t* e[])
+s8 calc_navigation_measurement(u8 n_channels, const channel_measurement_t *meas[],
```
jacobmcnamee: I'd say add some comments on the return values, but https://github.com/swift-nav/libswiftnav/pull/334 already adds doxygen for the whole thing, so I suppose this is fine here.

jacobmcnamee: Maybe just add 
```
/**
 * \return  0 on success,
 *         -1 if ephemeris is invalid
 */

**Comment by jacobmcnamee** [comment]
`calc_sat_state()` is also called in `make_propagated_sdiffs()` in observation.c. I think something like
```
      if (calc_sat_state(e[i], t, local_sat_pos, local_sat_vel,
                         &clock_err, &clock_rate_err) != 0) {
        continue;
      }
``` 
should work there.

jacobmcnamee:
lgtm

PR #346, opened 2016-05-12
Fix uninitialized ref_ecef

jacobmcnamee:
Fix passing unitialized `ref_ecef` to `update_ambiguity_test()` due to shadowing:
```
double ref_ecef[3];
{
  double ref_ecef[3];
  compute_ref_ecef(ref_ecef);
}
use_ref_ecef(ref_ecef);
```


ljbade:
Nice catch!

PR #347, opened 2016-05-12
Pass receiver time delta into calc_navigation_measurement()

jacobmcnamee:
Compute a receiver time delta for each measurement externally instead of passing in `nav_time` and computing it internally. Avoids subtracting two potentially large doubles to get a small number.


ljbade:
@jacobmcnamee Can you include the fix from https://github.com/swift-nav/libswiftnav/pull/334#discussion_r63763487 in this PR? since this is the one that changed it

/cc @scottswift

PR #358, opened 2016-05-27
lock_detect: minor fix in lock detector

valeri-atamaniouk:
Updated lock detector algorithm to avoid float to double conversions.


valeri-atamaniouk:
@jacobmcnamee Please review

ljbade:
Looks good to me. Also need to make PR for private.

PR #362, opened 2016-06-15
Add iono and tropo correction usage

dt-exafore:
This PR is created based on review here https://github.com/swift-nav/piksi_firmware/pull/686#discussion_r66972864
The idea is to move iono and tropo correction from FW to libswiftnav.
Since this is a public repository the function does not take into account L2C.

### DIFF
--- a/src/pvt.c
+++ b/src/pvt.c
@@ -598,3 +598,56 @@ s8 calc_PVT(const u8 n_used,
 
   return raim_flag;
 }
+
+/** Calculates Iono and Tropo correction
+ * \param n_ready_tdcp The number of measurements written to `m_tdcp`
+ * \param nav_meas_tdcp Pointer to measurements array
+ * \param pos_ecef position in WGS84 Earth Centered, Earth Fixed (ECEF) Cartesian
+ * \param pos_llh position in WGS84 geodetic coordinates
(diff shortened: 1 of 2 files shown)


dt-exafore:
@ljbade, @nsirola Please, review.

ljbade:
Sorry for the delay will look at it today.

src/pvt.c
```diff
@@ -598,3 +598,56 @@ s8 calc_PVT(const u8 n_used,
...
+/** Calculates Iono and Tropo correction
+ * \param n_ready_tdcp The number of measurements written to `m_tdcp`
+ * \param nav_meas_tdcp Pointer to measurements array
+ * \param pos_ecef position in WGS84 Earth Centered, Earth Fixed (ECEF) Cartesian
+ * \param pos_llh position in WGS84 geodetic coordinates
+ * \param iono_params pointer to ionosphere parameters received from L1CA stream */
+void calc_iono_tropo(const u8 n_ready_tdcp,
+                     navigation_measurement_t *nav_meas_tdcp,
+                     const double *pos_ecef,
+                     const double *pos_llh,
```
ljbade: I wonder if we should just pass in one form of the position, then convert it inside the function to the other?

